PR #133, opened 2024-02-01
STONEBLD-2133 Update pooled hosts

stuartwdouglas:


deploy/operator/update-host.yaml
```diff
@@ -0,0 +1,31 @@
...
+        #!/bin/bash
+        cd /tmp
+        set -o verbose
+        set -eu
+        set -o pipefail
+        mkdir -p /root/.ssh
+        cp $(workspaces.ssh.path)/id_rsa /tmp/master_key
+        chmod 0400 /tmp/master_key
+        export SSH_HOST=$(params.USER)@$(params.HOST)
+        ssh -i /tmp/master_key -o StrictHostKeyChecking=no $SSH_HOST "sudo dnf update -y"
```
arewm, line 30: Is the created user a sudoer? I would assume that they are not usually.

stuartwdouglas: This is running under the primary user (i.e. the user that creates the other users)

arewm (approved):

PR #134, opened 2024-02-02
Ephemerial Pools, first attempt

stuartwdouglas:


pkg/aws/ec2.go
```diff
@@ -130,19 +130,25 @@ func (configMapInfo AwsDynamicConfig) GetInstanceAddress(kubeClient client.Clien
...
 
-				return *res.Reservations[0].Instances[0].PublicDnsName, nil
-			}
+		server, _ := net.ResolveTCPAddr("tcp", *instance.PublicDnsName+":22")
+		conn, err := net.DialTCP("tcp", nil, server)
+		if err != nil {
+			log.Error(err, "failed to connect to AWS instance")
+			return "", err
 		}
+		defer conn.Close()
```
arewm, line 149: Is there no connection to close if the error is non-nil?

pkg/reconciler/taskrun/dynamicpool.go
```diff
@@ -0,0 +1,144 @@
...
+func (a DynamicHostPool) isHostIdle(r *ReconcileTaskRun, ctx context.Context, selectedHost string) (bool, error) {
+	trs := v1.TaskRunList{}
+	err := r.client.List(ctx, &trs, client.MatchingLabels{AssignedHost: selectedHost})
+	if err != nil {
+		return false, err
+	}
+	return len(trs.Items) == 0, nil
+}
+
+func (a DynamicHostPool) Allocate(r *ReconcileTaskRun, ctx context.Context, log *logr.Logger, tr *v1.TaskRun, secretName string, instanceTag string) (reconcile.Result, error) {
```
arewm, line 95: When is this called? When a new TaskRun is created that requests a dynamic provision? Will the allocated VM just go to the back of the queue and the one from the front of the queue pop off?

pkg/ibm/ibmz.go
```diff
@@ -129,6 +129,35 @@ func (r IBMZDynamicConfig) CountInstances(kubeClient client.Client, log *logr.Lo
...
+	if err != nil {
+		return nil, err
+	}
+	instances, _, err := vpcService.ListInstances(&vpcv1.ListInstancesOptions{ResourceGroupID: vpc.ResourceGroup.ID, VPCName: &r.Vpc})
+	if err != nil {
+		return nil, err
+	}
+	ret := []cloud.CloudVMInstance{}
+	for _, instance := range instances.Instances {
+		if strings.HasPrefix(*instance.Name, instanceTag) {
```
arewm, line 148: Is this instanceTag defined in the multi-platform deployment on each cluster itself so that we can identify the instances in a pool uniquely between all platforms? 

I think that it is possible to _label_ servers in IBM Cloud. Is leveraging these much more complex than just identifying a pattern on the instance name?

pkg/aws/ec2.go
stuartwdouglas: no, if you get an error you don't get a connection

pkg/ibm/ibmz.go
stuartwdouglas: For IBM I just use it as part of the instance name.

pkg/reconciler/taskrun/dynamicpool.go
stuartwdouglas: This is called when a task requests a VM. There is no queue as such, hosts are allocated based on current load. 

This implementation is kinda stateless, in that the state is maintained by either the cloud provider or the cluster.

PR #206, opened 2024-04-24
Do not fail AWS EC2 provision on connectivity check failure

mmorhun:
Recently, we've seen issues with setting up AWS EC2 instances for ARM builds.
Currently, provision is terminated on the first connectivity check to the VM.
Make connectivity check a transient error, so it will retry and fail only if we reach VM provision timeout.

### DIFF
--- a/pkg/aws/ec2.go
+++ b/pkg/aws/ec2.go
@@ -3,6 +3,10 @@ package aws
 import (
 	"context"
 	"fmt"
+	"net"
+	"os"
+	"strconv"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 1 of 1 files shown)


pkg/aws/ec2.go
```diff
@@ -138,7 +139,12 @@ func (r AwsDynamicConfig) GetInstanceAddress(kubeClient client.Client, log *logr
 	if len(res.Reservations) > 0 {
 		if len(res.Reservations[0].Instances) > 0 {
 			instance := res.Reservations[0].Instances[0]
-			return r.checkInstanceConnectivity(&instance, log)
+			address, err := r.checkInstanceConnectivity(&instance, log)
+			if err != nil {
+				// this might be transient, wait more for the instance to be ready
+				return "", nil
+			}
+			return address, nil
```
hugares, line 147: The only problem I see with the code now is considering all connectivity problems as transient might hide permanent connectivity issue and it will only fail when allocation time out is reached(which is better than the current issue IMHO).

That said, I am not knowledgeable enough of the code to be able to propose something else sot let's get this change in to hopefully unblock arm64 builds on prod and @stuartwdouglas can review this PR post merge

stuartwdouglas: Yea, this was a mistake on my part, I had an earlier PR where I was trying to stop the situation that @hugares is describing, where failed nodes would just keep being checked until the task timed out, but this error should have always been considered transient.

PR #265, opened 2024-07-16
Remove generate tasks

stuartwdouglas:
These are not actually used, and where from back when an early prototype provided its own CRDs.


ifireball (approved):
Ah I love the smell of code being deleted in the morning :)

/lgtm

PR #276, opened 2024-08-01
Makefile improve & automate RBAC generation

mshaposhnik:
Some makefile improvements  and reqiured roles autogeneration


ifireball (approved):
LGTM, but I wonder of we would get this if we migrate to a newer version of the KubeBuilder scaffolding.

Setting hold, as I wouldn't merge this before redhat-appstudio/infra-deployments#4241 is merged. I would probably also want [KFLUXINFRA-690](https://issues.redhat.com/browse/KFLUXINFRA-690) and [KFLUXINFRA-692](https://issues.redhat.com/browse/KFLUXINFRA-692) imlemented so we can verify our functionality in STAGE before pushing to PROD.

PR #277, opened 2024-08-01
feat(KFLUXINFRA-741): Add "Local" cloud

ifireball:
The "Local" cloud configuration is meant to allow a remote-capable
piepline task to run locally in the cluster by creating a dummy SSH
secret for it that only has the "host" key set to "localhost.

While we could in theory create a host pool that contains "localhost",
that would've caused the controller to try to SSH into it to configure
user accounts, etc.


ifireball:
This helps complete the picture for konflux-ci/build-definitions#1216

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -650,6 +701,17 @@ func createDynamicPoolHostConfig() []runtimeclient.Object {
 	return []runtimeclient.Object{&cm, &sec}
 }
 
+func createLocalHostConfig() []runtimeclient.Object {
+	cm := v1.ConfigMap{}
+	cm.Name = HostConfig
+	cm.Namespace = systemNamespace
+	cm.Labels = map[string]string{ConfigMapLabel: "hosts"}
+	cm.Data = map[string]string{
+		"local-platforms": "linux/arm64",
```
mshaposhnik, line 710: let's use constant here if possible

ifireball: I tent to avoid that in tests to make sure I test the actual values and not constants that might be set to the wrong values.

mshaposhnik: ok

mshaposhnik (approved):

pkg/reconciler/taskrun/taskrun_test.go
arewm, line 710: This mapping for a local-platform is only for the test, right? We may choose to use a similar pattern for local, but we may define any value that we want? For example, I had considered `local/arm64` but am okay to change if needed.

pkg/reconciler/taskrun/local.go
```diff
@@ -0,0 +1,57 @@
...
+		"host": []byte("localhost"),
+	})
+	return reconcile.Result{}, err
+}
+
+func (l Local) Deallocate(r *ReconcileTaskRun, ctx context.Context, tr *pipelinev1.TaskRun, secretName string, selectedHost string) error {
+	return nil
+}
+
+func createUserTaskSecret(r *ReconcileTaskRun, ctx context.Context, tr *pipelinev1.TaskRun, secretName string, secretData map[string][]byte) error {
```
arewm, line 38: Is this the function that is needed for the OTP functionality?

arewm (approved):

pkg/reconciler/taskrun/local.go
```diff
@@ -0,0 +1,57 @@
...
+	log.Info("Task set to run locally in the cluster")
+	tr.Labels[AssignedHost] = "localhost"
+	controllerutil.AddFinalizer(tr, PipelineFinalizer)
+	if err = r.client.Update(ctx, tr); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = createUserTaskSecret(r, ctx, tr, secretName, map[string][]byte{
+		"host": []byte("localhost"),
+	})
```
arewm, line 30: This is a convention that we are defining. Should we document this in the repo itself? Or will we just document it elsewhere?

pkg/reconciler/taskrun/local.go
ifireball: No, there is no OTP involved for the "local" hosts. This is for creating the secret the user task will mount. For actual remote hosts the secret is indeed generated via the OTP service, but that is done with a shell script running in a Task elsewhere.

pkg/reconciler/taskrun/taskrun_test.go
ifireball: Yes, this is test data.

I don't think it would be wise for us to call it "local/anything", wither an architecture runs in a "local pod" or a "remote VM" is not something that we should explicitly expose to user pipeline code. 
Consider what will happen when we will migrate from amd64 to arm64 for our production clusters, the "local" architecture will change, but we wouldn't want all the users to have to change their pipelines to use the new clusters.

ifireball:
Setting hold, please do not merge until redhat-appstudio/infra-deployments#4241 is merged.

PR #283, opened 2024-08-14
fix(KFLUXINFRA-784)  Ibmp dynamic pool ability

mshaposhnik:
- Implements functions needed for  dynamic-pool Pover VM
 - Adds power-go client which simplifies the interactions with PowerPC API
 
 Resolves: https://issues.redhat.com/browse/KFLUXINFRA-784


ifireball:
please clean up this PR - rebase the actual code change commit and remove all the unrelated updates

pkg/ibm/ibmp.go
```diff
@@ -140,7 +108,60 @@ func (r IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
 }
 
 func (r IBMPowerDynamicConfig) ListInstances(kubeClient client.Client, ctx context.Context, instanceTag string) ([]cloud.CloudVMInstance, error) {
-	return nil, fmt.Errorf("not impelemented")
+	log := logr.FromContextOrDiscard(ctx)
+	service, err := r.authenticate(kubeClient, ctx)
+	if err != nil {
+		return nil, err
+	}
+	builder := core.NewRequestBuilder(core.GET).WithContext(ctx)
```
ifireball, line 116: can we move the request building mechanics to different functions or even different abstraction objects so that the code here clearly show the logic of what we're actually doing?

pkg/ibm/ibmp.go
```diff
@@ -140,7 +108,60 @@ func (r IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
...
+	}
+
+	var rawResponse map[string]json.RawMessage
+	_, err = service.Request(request, &rawResponse)
+	if err != nil {
+		return nil, err
+	}
+	instancesData := rawResponse["pvmInstances"]
+	var instances []json.RawMessage
+	err = json.Unmarshal(instancesData, &instances)
```
ifireball, line 141: can we unmarshal directly into a slice of structures so that the structure of the data we're getting becomes clearer and we can possibly avoid or at lease simplify the translation loop below? 

Actually we probably want to do a little re-factor here to make CloudVMInstance be an interface instead of a plain struct, so that we can have a cloud-specific implementation for it here that we can directly unmarshal into.

mshaposhnik: I am currently experimenting with using 
https://github.com/IBM-Cloud/ppc-aas-go-client/tree/main
so far we can use just the model, and later re-done the whole PPC provider to use it (that's a good piece of work but imo it worth it)

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/konflux-ci/multi-platform-controller
 go 1.20
 
 require (
+	github.com/IBM-Cloud/power-go-client v1.6.0
```
hugares, line 6: Question so I can learn: if we use go modules, why are we adding the vendored dependencies to git?

mshaposhnik: thats builds faster and safer, but adds a lot of pain in ___ if you try to add something (like this PR)
maybe @ifireball  knows the historical background....

hugares: thanks for the explanation

ifireball: I'd drop that, but not a priority ATM.

pkg/ibm/ibmp.go
```diff
@@ -140,7 +119,55 @@ func (r IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
...
+		return nil, err
+	}
+
+	ret := make([]cloud.CloudVMInstance, 0, len(instances.PvmInstances))
+	for _, instance := range instances.PvmInstances {
+		if !strings.HasPrefix(*instance.ServerName, instanceTag) {
+			continue
+		}
+		identifier := cloud.InstanceIdentifier(*instance.PvmInstanceID)
+		createdAt, err := time.Parse(time.RFC3339Nano, instance.CreationDate.String())
```
ifireball, line 158: isn't CreationDate already a time value? If we need to do this parsing here, that models library is not doing much for us at all...

pkg/ibm/ibmp.go
```diff
@@ -140,7 +119,55 @@ func (r IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
...
+	}
+	builder.AddHeader("CRN", r.CRN)
+	builder.AddHeader("Accept", "application/json")
+
+	request, err := builder.Build()
+	if err != nil {
+		return nil, err
+	}
+
+	instances := models.PVMInstances{}
```
ifireball, line 146: its does not seem like using that models library provides a net improvement for the code

pkg/ibm/ibmp.go
mshaposhnik: yep, its an wrapper wich can be just cast to `Time`, fixed

pkg/ibm/ibmp.go
mshaposhnik: Its already sets us free from those manual json manipulation mess and/or writing custom models.
With some more investment it will allow making the API requests out of the box without having to manuallly constructing them.

pkg/ibm/ibmp.go
```diff
@@ -155,10 +155,7 @@ func (r IBMPowerDynamicConfig) ListInstances(kubeClient client.Client, ctx conte
 			continue
 		}
 		identifier := cloud.InstanceIdentifier(*instance.PvmInstanceID)
-		createdAt, err := time.Parse(time.RFC3339Nano, instance.CreationDate.String())
-		if err != nil {
-			log.Error(err, "cannot parse creation time", "instance", identifier)
-		}
+		createdAt := time.Time(instance.CreationDate)
```
ifireball, line 158: is this a type cast? why do we need it? What is the type on CreationDate?

mshaposhnik: it is [strfmt.DateTime](https://github.com/go-openapi/strfmt/blob/master/time.go#L125) which is json utility wrapper around `time.Time`

ifireball: ok, thanks.

ifireball (changes requested):
LGTM, but please cleanup the commits in this PR - make sure all are based on a recent `main` and squash fix/junk commits into the main commits that actually do the desired change.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -469,9 +469,7 @@ func extracPlatform(tr *tektonapi.TaskRun) (string, error) {
 
 func (r *ReconcileTaskRun) handleHostAllocation(ctx context.Context, tr *tektonapi.TaskRun, secretName string, targetPlatform string) (reconcile.Result, error) {
 	log := logr.FromContextOrDiscard(ctx)
-
-	log.Info("attempting to allocate host")
-
+	log.Info("attempting to allocate host for platform:" + targetPlatform)
```
ifireball, line 472: Please use logr key/values instead of embedding values into the string
```suggestion
	log.Info("attempting to allocate host", "platform", targetPlatform)
```
Also, are you sure we don't already embed the platform value into the logger somewhere? We probably should once we know what the requested platform is, so it gets added to all logged lines

pkg/ibm/ibmp.go
```diff
@@ -140,7 +119,52 @@ func (r IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
 }
 
 func (r IBMPowerDynamicConfig) ListInstances(kubeClient client.Client, ctx context.Context, instanceTag string) ([]cloud.CloudVMInstance, error) {
-	return nil, fmt.Errorf("not impelemented")
+	log := logr.FromContextOrDiscard(ctx)
+	log.Info("Listing instances by tag:" + instanceTag)
```
ifireball, line 123: use structured log values instead of embedding into the string
```suggestion
	log.Info("Listing instances", "tag", instanceTag)
```

pkg/reconciler/taskrun/taskrun.go
mshaposhnik: No, practically we have only request name and namespace:
`{"request": {"name":"mpc-stg-test-repo-on-push-xh46v-build-images-1","namespace":"mshaposh-tenant"}}`
I'll commit your suggestion.

ifireball (approved):

PR #295, opened 2024-09-13
refactor metrics as prerequisite to adding availability probes

mshaposhnik:


mshaposhnik:
@ifireball WDYT ?

pkg/metrics/common.go
```diff
@@ -0,0 +1,23 @@
+package mpcmetrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+const (
+	MetricsSubsystem = "multi_platform_controller"
+	MetricsNamespace = "redhat_appstudio"
```
hugares, line 7: Any reason why not using new org name `konflux_ci` instead of  `redhat_appstudio` ?

ifireball:
> @ifireball WDYT ?

please help me understand what are the issue with the existing metrics code that we are trying to fix here.

mshaposhnik:
Existing code is:
 - only provides per-platform metrics
 - mixed up with reconciler code, thus completely hard to support/debug
 - untested

pkg/metrics/common.go
mshaposhnik: That is seems to be still default namespace, as far as I can see from other components.
I'll check with `o11y` guys it that really makes some difference.

pkg/metrics/common_test.go
```diff
@@ -0,0 +1,37 @@
+package mpcmetrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	prometheusTest "github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestRegisterCommonMetrics(t *testing.T) {
```
ifireball, line 10: @mshaposhnik I know that other tests in this repo were not written using Ginkgo - but can you make an effort to use it for new tests?

Also please try to use Gomega rather then testify for assertions

pkg/metrics/common.go
ifireball: @mshaposhnik lets get this conversation resolved and then we can merge this

ifireball (approved):

pkg/metrics/common.go
mshaposhnik: After looking at o11y metrics  rules, it seems that namespace is not necessary, i will leave only namespace.
Aviablity metric will have value `multi_platform_controller_available` which is imo self-descriptive enough.

PR #297, opened 2024-09-16
fix(KFLUXINFRA-836): Bump tekton to v0.62.0

ifireball:
Looks like there is an additional deprecated feature flag that was
removed.


hugares (approved):
/lgtm

PR #309, opened 2024-10-10
Avialability metrics introduction

mshaposhnik:
This PR introduces simple per-provider availability probe checker and metric exposer.
The  evaluation logic is following:
 - we count VM provision successes and failures separately per each platform for 5 minutes;
 - each 5 minutes counters are evaluated and reset. 

The platform liveness consideration takes place according to the following rules:
- No sucesses, no errors - no platform liveness data reported.
 - Zero successful provisioned VMs, but non-zero failures - platform considered as failed;
 - Non-zero successes - errors to success rate is calculated. If errors  portion is less than 50%, platform considered healthy, otherwise failed.
 
 Example of exposed platform liveness metric:
 
Screenshot_20241017_102632


pkg/metrics/backend_probe.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func NewBackendProbe() AvailabilityProbe {
+	return &BackendProbe{
+		successes: atomic.Int32{},
+		failures:  atomic.Int32{},
+	}
+}
+
+func (q *BackendProbe) CheckAvailability(_ context.Context) error {
```
ifireball, line 23: I wonder if we can emit simpler metrics and do this calculation on the Prometheus side with recording rules

mshaposhnik: It is possible theoretically, but that might be not so reliable, as the transformer might become complex and need tweeks if some metrics logic is changed (or supplying unreliable data). So i decided to count that explicitly. And that might also allow us to add more health checks whih are not covered by regular metric counters. I did follo the general recoomendations here:

ifireball: It might indeed be useful to bring @yftacherzog and @amisstea into this discussion....

WDYT guys? Should we do this kind of processing in the code or should we use recording rules?

yftacherzog: I agree that it mainly depends on the sort of data that needs to be manipulated.

ifireball: ok, this discussion isn't leading anywhere useful it seems so I'm gonna close it.

ifireball (approved):

PR #311, opened 2024-10-11
Add reschedule after obtaining instance ID

mshaposhnik:
Added rescheduling of TaskRun reconsilliation after obtaining instance ID on dynamic platform, to avoid timing outs.
Fixes https://issues.redhat.com/browse/KFLUXINFRA-932

### DIFF
--- a/pkg/reconciler/taskrun/dynamic.go
+++ b/pkg/reconciler/taskrun/dynamic.go
@@ -228,6 +228,6 @@ func (r DynamicResolver) Allocate(taskRun *ReconcileTaskRun, ctx context.Context
 		}
 	}
 
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: 2 * time.Minute}, nil
 
 }


pkg/reconciler/taskrun/dynamic.go
```diff
@@ -228,6 +228,6 @@ func (r DynamicResolver) Allocate(taskRun *ReconcileTaskRun, ctx context.Context
 		}
 	}
 
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: 2 * time.Minute}, nil
```
hugares, line 231: Can you expand on what this changes does. Does this means that we will wait for an extra 2 minutes before letting the provision task run?

mshaposhnik: Hi, not exactly. 
If `requeue` is not set, that means next reconcillitation will happen _only_ when Taskrun will be modified.
That's why we're missing reaction to VM become available for the Taskun, because just nothing veriifies this fact, as nothing changes the TaskRun 
Setting this time out forces reconcile the object just by passing some time. 2 min is empiric estimated value which we may have found VM ready from the first try, and not too much overhead if not.

hugares: thanks for the explanation

hugares (approved):
/lgtm

brianwcook:
@hugares can you verify when this is deployed, we need to know when to test it

hugares:
@brianwcook Change is being rolled out to prod now, @mshaposhnik Can you confirm it is in prod now?

mshaposhnik:
Yes itt is now rolled, i'll keep an eye on how things will go this evening

PR #314, opened 2024-10-17
Make metrics endpoint secure using kube-rbac proxy

mshaposhnik:
Currently MPC metrics endpoint is unsecrure, which is the potential security gap.
Ths PR make metrics secured usin standard `kube-rbacproxy` sidecar approach.


hugares (approved):
/lgtm

PR #325, opened 2024-11-01
Use VNI attachment instead of network in ibmz

mshaposhnik:
To get VPN working we need VNI attachment instead of simple network
See https://issues.redhat.com/browse/KFLUXINFRA-931


hugares:
LGTM
we need to make sure this works in both type of IBM account we have, i.e. internal and external

bamachrn (approved):
LGTM

bamachrn (approved):

ifireball (approved):

PR #326, opened 2024-11-01
Set the explicit timeout for TCP address probe

mshaposhnik:
Sometimes it is visible from logs that TCP port probe we're using to check VM is responsible may take up to 2 minutes, blocking the execution thread. This PR sets explicit 5 sec timeout for TCP probe.


hugares (approved):
/lgtm

ifireball (approved):

PR #336, opened 2024-11-07
Temporary fix to norrow the scope of cached TaskRuns to prevent OOMs

mshaposhnik:
We have OOMs on prod,  with mpc container consuming over 8Gb RAM. This is supposedly due ti high number of TaskRuns on cluster (>10000, and we cache them all)
This PR limits caches TRs to just those belonging to MPC

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -29,6 +29,7 @@ var (
 )
 
 const TaskRunLabel = "tekton.dev/taskRun"
+const TaskLabel = "tekton.dev/task"
 
 func NewManager(cfg *rest.Config, options ctrl.Options) (ctrl.Manager, error) {
 	// do not check tekton in kcp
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -81,9 +82,17 @@ func NewManager(cfg *rest.Config, options ctrl.Options) (ctrl.Manager, error) {
 		return nil, lerr
 	}
 	podSelector = podSelector.Add(*podLabels)
+
+	taskrunSelector := labels.NewSelector()
+	trLabels, lerr := labels.NewRequirement(TaskLabel, selection.In, []string{"buildah-remote-oci-ta", "clean-shared-host", "provision-shared-host", "update-host"})
```
hugares, line 87: humm, I have a concern about this change. First, there are more task to watch than buildah-remote-oci-ta, there is the regular one and all the various flavors of buildah with larger resources.

The concern is everytime a task will be added, MPC will not react until we update this.

mshaposhnik: I agree, this proposal far from ideal. But unfortunately we have no other options to filter the TaskRun-s multiplicity, other than by labels or some common fields(like `metadata.Name` and `metadata.Namespace`).
 
As far as i was able to understood from docs, caching of objects does not affects ability of their processing by controller (and this is proofed by the fact that most konflux controlles does not use caching at all), it is just means they're always will be read directly from etcd on `Get` operations which is increases the api load and slowing down the reconcillation speed.

ifireball (commented):
I'm trying to understand what would be the downside of doing this. Worst case - if we need to reconcile a task that is not in the cache, then we end up with slower reconciliation for it?

filariow:
To further improve memory usage, we can also consider using the [TransformFunc](https://github.com/kubernetes-sigs/controller-runtime/blob/e08f24a1e29fb84e4b059dd471b1c8b4928751e7/pkg/cache/cache.go#L286) to [trim resources](https://github.com/kubernetes/client-go/blob/ec5b832e6085a572b8c325dafe9323ff18174732/tools/cache/delta_fifo.go#L150-L151). Like removing the `managedFields`, `ownerReferences`, unneeded `annotations`, etc.

PR #343, opened 2024-11-17
feat(KFLUXINFRA-930): Error secrets: More CTX info

ifireball:
Add more contextual information to error secrets.


meyrevived:
One checks still fails at re-run and I don't have access to rhtap-build-tenant to see what it is that failed.

mshaposhnik (approved):

ifireball:
> One checks still fails at re-run and I don't have access to rhtap-build-tenant to see what it is that failed.

See here about that check: https://issues.redhat.com/browse/KFLUXSPRT-419

PR #344, opened 2024-11-27
Add debug level logging and align the messages between platforms

mshaposhnik:
Added ability to configure the logger level on startup and align/fixup some messages

### DIFF
--- a/pkg/ibm/ibmp.go
+++ b/pkg/ibm/ibmp.go
@@ -55,6 +55,8 @@ func IBMPowerProvider(platform string, config map[string]string, systemNamespace
 }
 
 func (r IBMPowerDynamicConfig) LaunchInstance(kubeClient client.Client, ctx context.Context, taskRunName string, instanceTag string, _ map[string]string) (cloud.InstanceIdentifier, error) {
+	log := logr.FromContextOrDiscard(ctx)
+	log.Info(fmt.Sprintf("attempting to launch IBM-P instance for %s", taskRunName))
 	service, err := r.authenticatedService(ctx, kubeClient)
 	if err != nil {
 		return "", err
@@ -74,6 +76,8 @@ func (r IBMPowerDynamicConfig) LaunchInstance(kubeClient client.Client, ctx cont
 }
 
 func (r IBMPowerDynamicConfig) CountInstances(kubeClient client.Client, ctx context.Context, instanceTag string) (int, error) {
+	log := logr.FromContextOrDiscard(ctx)
+	log.Info("attempting to count IBM-P instances")
 	instances, err := r.fetchInstances(ctx, kubeClient)
 	if err != nil {
 		return 0, err
@@ -84,6 +88,7 @@ func (r IBMPowerDynamicConfig) CountInstances(kubeClient client.Client, ctx cont
 			count--
 		}
 	}
+	log.Info("Count of IBM-P instances done", "count", count)
 	return count, nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/logs/logs.go
```diff
@@ -0,0 +1,39 @@
...
+	zap2 "go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const DebugLevel = 1
+
+func InitLogger(logLevel, stackTraceLevel string) error {
```
ifireball, line 14: I don't think we need any of this - to log at a higer (less siginificant) level, all we need to do is `log.V(1).Info(...)` with a normal logr object.

pkg/ibm/ibmp.go
```diff
@@ -114,26 +119,26 @@ func (r IBMPowerDynamicConfig) authenticatedService(ctx context.Context, kubeCli
...
-	log := logr.FromContextOrDiscard(ctx)
+	log := logr.FromContextOrDiscard(ctx).WithValues("platform", "ibm-p", "instanceId", instanceId)
 	service, err := r.authenticatedService(ctx, kubeClient)
 	if err != nil {
 		return "", err
 	}
 	ip, err := r.lookupIp(ctx, service, string(instanceId))
 	if err != nil {
-		log.Error(err, "Failed to lookup IP", "instanceId", instanceId, "error", err.Error())
+		log.Info("Failed to lookup IP", "error", err.Error())
```
ifireball, line 129: IMO we should keep logging errors with `Error`

pkg/ibm/ibmp.go
```diff
@@ -114,26 +119,26 @@ func (r IBMPowerDynamicConfig) authenticatedService(ctx context.Context, kubeCli
...
+		log.Info("Failed to check address", "error", err.Error())
 		return "", nil
 	}
 	return ip, nil
 }
 
 func (r IBMPowerDynamicConfig) ListInstances(kubeClient client.Client, ctx context.Context, instanceTag string) ([]cloud.CloudVMInstance, error) {
-	log := logr.FromContextOrDiscard(ctx)
-	log.Info("Listing ppc instances", "tag", instanceTag)
+	log := logr.FromContextOrDiscard(ctx).WithValues("platform", "ibm-p")
```
ifireball, line 140: the `platform` keywork should probably be reserved to the `PLATFORM` parameter value and we should set it as soon as we know it and ensure its inherited by all logs from that point on.

The value here is probably better named "driver" or something else that is consistent with how we call these things in the code.

ifireball: yeah, I just forget what the code calls those.

pkg/ibm/ibmp.go
mshaposhnik: ok

pkg/logs/logs.go
mshaposhnik: Sorry I didn't get the message. This code allows to start the controller with any desired logging level. I.e. by default it will be `info` (so no debugs), but using `--zap-log-level=debug` (or any other valig level) it can be switched on.

ifireball: I think the built-in zap options aught to be enough. Also I don't think this aught to be very configurable. I don't want to be unable to debug a production issue because the messages I need happen to be in a level we were not tracking. 

In other words, IMO level filtering should be done when querying at the destination (e.g. Splunk) and not at the source.

mshaposhnik: I would say it is not any easier to debug the issue when log is overpolluted by meaningless messages like `pipelinerun xxxx was probably deleted`  et cetera. They doesnt carry any payload, just the log size grows enormous. The idea was to mute such unifromative stuff by def.

ifireball: if you really think a message is unnecessary, remove it from the code, otherwise you can never know what you'll need to debug stuff for users.

mshaposhnik: OK, so what should we leave in the PR then? To me it is have not much meaning move the log levels without filtering ability.

mshaposhnik:
Closing,  as without the ability to change level this will not make MPC logs any better.
We will still see at most the last ~2 hrs of the controller log,  due to it's overpolluting and therefore log being truncated by Openshift. If one needs some history more than this, it will need to make it via Splunk.

PR #351, opened 2024-12-09
Expose events for a few transitions

ralphbean:
I found that when MPC fails, the user has very little info in their namespace about why. If we expose events via the kube events API, then the users can better help us triage issues.


mshaposhnik (approved):

ifireball (approved):
I actually have an implementation of a logger that can emit events in another controller, using that would have been nicer, but I guess I can live wth this,

PR #354, opened 2024-12-12
Distinguish recorded events from tekton

ralphbean:
When reviewing `kubectl get events`, it is difficult to tell which events come from MPC and which events come from tekton itself.

Using a different term here will make things more obvious to users trying to work with our events to understand what MPC is doing.


ralphbean:
For example, compare this with the multiple different event recorders used by build-service: https://github.com/konflux-ci/build-service/blob/main/main.go#L165

ifireball (approved):

PR #355, opened 2024-12-12
Tweak event and log language

ralphbean:
This is just a little confusing. There are two parts, launching the VM and provisioning the VM. Here, the provisioning logline said that it was "launching" the provisioning task which, while correct, made it a little confusing when reading through multiple log lines and multiple events.

Just change the word to make it clear that it's a separate phase, not more "launching".


hugares:
LGTM

ifireball (approved):

PR #361, opened 2024-12-18
Add local development scripts

mshaposhnik:
This PR allows to deploy MPC to any kind of locally running kubernetes, i.e. minikube or k3s or whatever.
It is _not_ yet tested with AWS and IBM cloud credentials, that will be done separately.

### DIFF
--- a/deploy/generate_ca.sh
+++ b/deploy/generate_ca.sh
@@ -19,5 +19,8 @@ openssl req -x509 -new -nodes -key ${GENCERTS_DIR}/ca.key -subj "/CN=otp-service
 # verify created
 cat ${GENCERTS_DIR}/ca.crt >/dev/null
 if [ $? -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


deploy/generate_ca.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
ifireball, line 1: the Kube builder documentation talks about using cert-manager for this sort of thing, can't we do the same?

I don't like assuming the openssl CLI is installed locally, That is not always the case.

mshaposhnik: Installing cert-manager with configuring all the necessary stuff  such as Issuers, CertificateRequests etc to have just one cert generated is a huge overkill for me. I can hardly imagine an developer machine w/out having `openssl`.

ifireball (approved):

PR #360, opened 2024-12-18
Integration KFLUXINFRA-1143 - refining the post-merge SeaLights scan on merged main (no test runs)

meyrevived:
Dealing with PR contexts again because of SeaLights need for an agent token. Because the post-merge action (pull_request on a closed pull request) is still running in the context of the PR branch (the forked repository that has no access to the SeaLights token stored as a repository token on the main repository) this workflow needs to be changed to just be a stub workflow that triggers a workflow_run (committed now) that will run only if it's in the context of the main repository.

In this PR:
1. Changes to the workflow that runs a SeaLights scan on a merged main so that it won't be triggered on its own.
2. Removal of the extra log writes for testing SeaLights noticing code changes, so that Eran's [PR](https://gitlab.cee.redhat.com/konflux/infra/-/merge_requests/297) can run be merged

PR #362, opened 2024-12-19
KFLUXINFRA-1143: SeaLights scan on merged main implementation via github workflows

meyrevived:
This pull request adds the (hopefully) final workflow that runs on a closed pull request and is an event the workflow_run workflow from pull request #360 that actually runs SeaLights scan on the new main branch, listens to.
Also in this PR is a small change to the if statement in the workflow_run file that is a more percise check to verify it's definitely not running on the forked branch context.


hugares (approved):
/lgtm

PR #363, opened 2024-12-19
Do not log error on 404 status on subsequent deletion repeats

mshaposhnik:
Poeple noticing too much `failed to delete system z instance, unable to get instance` messages in logs, while they can be part of a normal deletion verification process


hugares (approved):
/lgtm

ifireball (approved):

PR #364, opened 2024-12-19
KFLUXINFRA-1143: fix for wrong if condition in workflow_run 

meyrevived:
The workflow_run was getting skipped because I coded it to check the triggering event's repo, not the repo all of these actions were being run on. Because this workflow is triggered by a pull request, the event repo is never the main repo, which is not what I needed. Thus, the workflow was being skipped all the time.


hugares (approved):
/lgtm

PR #368, opened 2024-12-25
KFLUXINFRA-1064  adding checks for failed pull request workflow runs

meyrevived:
I've noticed that the [pull_request_target workflow](https://github.com/konflux-ci/multi-platform-controller/actions/workflows/sealights-mpc-test-ci.yaml) has been triggered by no apparent action and that the step for configuring the SeaLights agent there is failing because there's missing data from the workflow run.
To debug when this is happening, debug that this is definitely something in our system and not something SeaLights can't find and to prevent this workflow from running in parallel (thus leading to missing context information needed for the SeaLights agent) I've added some additional steps to this workflow.

Additionally, there are additions to go-ci.yaml to ensure the pull_request_target workflow that's running the unit tests is monitored for failure so that if it does fail, go-ci fails as well. This is because I did not see the [failed runs](https://github.com/konflux-ci/multi-platform-controller/actions/workflows/sealights-mpc-test-ci.yaml) being listed anywhere as failed checks for [the PR](https://github.com/konflux-ci/multi-platform-controller/pull/361) they were happening in parallel to.

### DIFF
--- a/.github/workflows/go-ci.yaml
+++ b/.github/workflows/go-ci.yaml
@@ -4,7 +4,7 @@ on:
     branches: [ main ]
 permissions:
   checks: read
-  contents: write
+  contents: read
   actions: write
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


meyrevived:
@ifireball?

mshaposhnik (approved):

PR #370, opened 2024-12-30
feat[KONFLUX-6145]:Make IBM volumes size configurable

mshaposhnik:
AWS-based platforms supports the volume size configuration, and IBM must do so. Some users requested that current default size in not enough.

IBM Z allows to specify requested disk size on VM startup, while IBM P requires either resizing volume after machine boot, or creating additional volume and attach it to VM, i chosen the 1st option so far.

Fixing: https://issues.redhat.com/browse/KONFLUX-6145


ifireball (approved):
lgtm

PR #371, opened 2024-12-30
KFLUXINFRA-1064: testing phase added to merged main branch SeaLights workflow and refactoring for DRY

meyrevived:
SeaLights asked that we'll add a testing phase to the github workflow that runs a SeaLights code scan on the main branch after it's been merged with new code.

This basically made sealights-mpc-test-ci.yaml extremely similar to this workflow and so I removed all the repeating code to a separate bash script that needs to be given flags depending on where it's run (pull request or merged main) and now the workflows just run this script with its flags.


mpc_testing_with_sealights.sh
```diff
@@ -0,0 +1,96 @@
...
+      ;;
+  esac
+done
+
+# Validate mode and required parameters
+if [[ -z "$MODE" ]]; then
+  echo "Error: --mode is required" >&2
+  exit 1
+fi
+
```
ifireball, line 45: I would separate downloading dependencies and running processes, so that GitHub Actions` cache can be used

meyrevived: SeaLights are working on fetching their agent via go releases currently and I'm checking in on their progress on a daily basis. Using github actions cache would still be identical in both workflows and wouldn't be DRY if I separated them. I'm not sure we should invest in a different method for the moment, if a different solution is about to come out any day now. SeaLights are based in Israel and have been working through this holiday season so progress is being made.

ifireball:
@meyrevived do we still need this PS given that #372 was merged?

meyrevived:
Made differently in https://github.com/konflux-ci/multi-platform-controller/pull/372

PR #372, opened 2024-12-31
KFLUXINFRA-1064: refactoring to run SeaLights on tests for pull request and merged main branch

meyrevived:
SeaLights asked that we'll add a testing phase to the github workflow that runs a SeaLights code scan on the main branch after it's been merged with new code.

This basically made sealights-mpc-test-ci.yaml extremely similar to this workflow and so I added code that runs this scan-and-test procedure to on->push events. Steps that need to run differently on pull requests and on push are checked to see how to run them. 
Thus the name change to sealights-mpc-test-ci.yaml, removing sealights-mpc-test-on-merged-main.yaml and trigger-sealights-mpc-test-on-merged-main.yaml. codecov-main.yaml is also removed because mpc-test-sealights.yaml is already running codecov action.


ifireball (approved):
nice, I like how we end up with less code fro more functionality.

PR #379, opened 2025-01-07
Use separate context for long running routines

mshaposhnik:
The common context object is cancelled when reconcilliation loop is over, so long running routines getting the `context is cancelled` error when making API requests  using `RequestBuilder`.

This is discovered on volume updates, but also may be a possible reason why machine cleanup doesn't work sometimes.

The proposed fix is to create local uncancellable copy of the parent context in long running routines.


ifireball (approved):
lgtm

PR #384, opened 2025-01-11
Do not fail resizing attemps on conflict error

mshaposhnik:


ifireball (approved):

ifireball:
@meyrevived not sure what the issue with SeaLights is, PTAL.

PR #397, opened 2025-01-20
Unit test for Ec2Provider 

meyrevived:
A new unit test to ec2.go -> Ec2Provider, which will be the only function covered by unit test from ec2.go once the [planned cloud provider API/functionality integration test](https://issues.redhat.com/browse/KFLUXINFRA-1065) is done and running.

This is a simple test that checks that arm64 and amd64 EC2 host configuration fields are filled, checked and corrected to better values properly - default values where they're needed, correct fields in the correct place etc.


mshaposhnik:
Can you explain which case we're testing with that enormous user data? Is expected behavior different than the normal one?

meyrevived:
Just some stress testing on a field that goes through a transformation. It
looks like userData is used for free form code typing, I just want to see
what happens if someone really pours in a lot of stuff in there.

mshaposhnik:
So if we don't testing any field size limits or anything else in this regard, i propose to maybe skip it, as it is does not add any value and just tests the `base64` library conversion, which is  surely cannot broke by this amount of data.

meyrevived:
Hmm, I do want to include these tests cases in the API integration test, because they are a challenging output downstream for the rest of the code that's providing a host. So I'll just comment them out.

PR #400, opened 2025-01-28
Add partition and filesystem resize for PPC custom volumes size

mshaposhnik:
Resizing the volume on-the-fly only resizes the disk, but partition and filesystem remain the same size.
They need to be aligned manually.
Those commands does not affect stardard size volumes (is no resize applied).
Fixes  https://issues.redhat.com/browse/KONFLUX-6560


meyrevived (approved):
/lgtm

PR #401, opened 2025-01-28
update konflux references

mshaposhnik:
task-buildah were migrated 0.2 to 0.3 in #395  but this is buggy version tag, need the fixed one


meyrevived (approved):
/lgtm

PR #411, opened 2025-02-14
Cloud Provider code cleanup

enkeefe00:
Several changes were made in the IBM and AWS cloud provider
packages to increase readability and decrease newcomer ramp-up time.
Comments were added per Go's style guide, variables were renamed,
functions were created etc.


mshaposhnik:
I'm afraid I'm not ready to approve this wihout being properly covered by tests and proven it is not introduces new bugs. This is the main reason why it wasn't refactored before - not because there are no people able to do this, but simply because without good coverage that may bring more problems than profit, imo. 
I'll left the final decision on @ifireball  :)

hugares:
> I'm afraid I'm not ready to approve this wihout being properly covered by tests and proven it is not introduces new bugs. This is the main reason why it wasn't refactored before - not because there are no people able to do this, but simply because without good coverage that may bring more problems than profit, imo. I'll left the final decision on @ifireball , though :)

Would you be more comfortable approving refactoring PR if the current PR would be broken down in smaller ones, scoped to one change or type of change per PR?

mshaposhnik:
Yes - definitely that might work better.  Like, configuration reading changes which are easy to test on stg. Then next part, etc.

enkeefe00:
> Yes - definitely that might work better. Like, configuration reading changes which are easy to test even on stg. Then next part, like ec2 changes etc.

I will break up this PR and the associated ticket. My apologies for creating such a large set of changes. :sweat_smile:

PR #420, opened 2025-02-20
IBM Power Systems Cloud Provider code cleanup (part 2)

enkeefe00:
* Variables were renamed for clarity
* More error messsages were prefaced with detail for clarity
* Non-CloudProvider interface functions were moved to a helper file

### DIFF
--- a/pkg/ibm/ibmp.go
+++ b/pkg/ibm/ibmp.go
@@ -11,26 +11,19 @@ import (
 	// strength doesn't matter here
 	"crypto/md5" //#nosec
 	"encoding/base64"
-	"encoding/json"
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
 	"time"
 
-	"github.com/IBM-Cloud/power-go-client/power/models"
-
-	"github.com/IBM/go-sdk-core/v5/core"
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
 	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
-	v1 "k8s.io/api/core/v1"
-	types2 "k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


meyrevived:
@enkeefe00  I wonder about some of the helper methods in this PR and the other two PRs.
pkg/cloud/cloud.go is an interface of all three cloud provider files and has functions that are supposed to interface for the functionalities, IBM-Z and EC2 provider code sticks to these names, but IBM-P doesn't (like [launchInstance](https://github.com/konflux-ci/multi-platform-controller/blob/d8251065e5d760841d46c2e227f8446a1c55f18b/pkg/cloud/cloud.go#L12) is supposed to be the interface for IBM-P's createServerInstance). 
Perhaps as part of the code cleanup, there could be an alignment with the interface names in IBM-P's code.

And while we're at it (in a different set of PRs, obviously) you can have a cloud_helper.go interface for the all three cloud provider's helper functions. What do you think? 
@mshaposhnik  what do you think?

enkeefe00:
> @enkeefe00 I wonder about some of the helper methods in this PR and the other two PRs. pkg/cloud/cloud.go is an interface of all three cloud provider files and has functions that are supposed to interface for the functionalities, IBM-Z and EC2 provider code sticks to these names, but IBM-P doesn't (like [launchInstance](https://github.com/konflux-ci/multi-platform-controller/blob/d8251065e5d760841d46c2e227f8446a1c55f18b/pkg/cloud/cloud.go#L12) is supposed to be the interface for IBM-P's createServerInstance). Perhaps as part of the code cleanup, there could be an alignment with the interface names in IBM-P's code.
> 
> And while we're at it (in a different set of PRs, obviously) you can have a cloud_helper.go interface for the all three cloud provider's helper functions. What do you think? @mshaposhnik what do you think?

My next set of PRs will be renaming/refactoring the helper functions. Thanks for the suggestion regarding the naming scheme! I'd be happy to incorporate that.

meyrevived (approved):
/lgtm

PR #428, opened 2025-03-04
Prevent VM names to exceed max allowed length

mshaposhnik:
Until tag length validation is added,  we must check max name length to avoid `400 BadRequest`
Fixes https://issues.redhat.com/browse/KFLUXINFRA-1328


enkeefe00 (approved):

meyrevived (approved):
I think these changes should stay here even after we (finally) have config validation. The config validation uses kyverno-cli, and  itcan mutate fields that fail validation if these fields are in an file instead of a resource in a k8s cluster.

PR #430, opened 2025-03-18
Fix conditional panic in MPC

enkeefe00:
When an instance cannot be retrieved while being deleted, the following panic occurs:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x80 pc=0x1df841a]
goroutine 1013 [running]:
github.com/konflux-ci/multi-platform-controller/pkg/ibm.IBMZDynamicConfig.TerminateInstance.func1()
/opt/app-root/src/pkg/ibm/ibmz.go:454 +0x3ba
created by github.com/konflux-ci/multi-platform-controller/pkg/ibm.IBMZDynamicConfig.TerminateInstance in goroutine 135
/opt/app-root/src/pkg/ibm/ibmz.go:446 +0x1dd
````
This nil pointer is no longer referenced.


enkeefe00:
@mshaposhnik @ifireball

mshaposhnik (approved):

PR #435, opened 2025-03-31
Add shell and unicode sast pipeline tasks

mshaposhnik:
For: https://issues.redhat.com/browse/KONFLUX-2264


.tekton/multi-platform-controller-otp-service-push.yaml
```diff
@@ -331,7 +412,7 @@ spec:
       - name: image-url
         value: $(tasks.build-container.results.IMAGE_URL)
```
enkeefe00, line 413: Does this need to be changed to 'build-image-index' too?

PR #438, opened 2025-03-31
Add functionality to re-queue failed AWS EC2s

enkeefe00:
Since MPC assumes a cloud provider's capacity is limitless, it will
create a virtual machine (VM) on the cloud provider and if the cloud
provider cannot start the VM due to a lack of capacity (which is
indicated by the VM being in a failed state), the user's build will fail.
This commit implements the cloud interface function necessary to
enter a re-queue process committed in PR https://github.com/konflux-ci/multi-platform-controller/pull/425.

### DIFF
--- a/pkg/aws/ec2.go
+++ b/pkg/aws/ec2.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"slices"
 	"strconv"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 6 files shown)


enkeefe00:
/hold

Needs to wait on PR#425 to be merged

pkg/aws/ec2.go
```diff
@@ -252,9 +252,32 @@ func (r AwsEc2DynamicConfig) SshUser() string {
 	return "ec2-user"
 }
 
-// TODO: implement this function.
+// GetState returns ec's VM state from AWS. See https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceState.html for
+// valid states.
 func (ec AwsEc2DynamicConfig) GetState(kubeClient client.Client, ctx context.Context, instanceId cloud.InstanceIdentifier) (string, error) {
-	return "ACTIVE", nil
+	log := logr.FromContextOrDiscard(ctx)
+	log.Info(fmt.Sprintf("Attempting to get AWS EC2 instance %s's IP address", instanceId))
```
filariow, line 259: alternatively, you can use the following format. It would easier to filter logs

```suggestion
	log.Info("Attempting to get AWS EC2 instance's IP address", "instanceId", instanceId)
```

pkg/aws/ec2.go
```diff
@@ -252,9 +252,32 @@ func (r AwsEc2DynamicConfig) SshUser() string {
...
+		// This might be a transient error, so only log it
+		log.Error(err, "failed to retrieve instance", "instanceId", instanceId)
+		return "", nil
+	}
+
+	// Get state
+	if len(instancesOutput.Reservations) > 0 {
+		if len(instancesOutput.Reservations[0].Instances) > 0 {
+			instance := instancesOutput.Reservations[0].Instances[0]
+			return string(instance.State.Name), nil
```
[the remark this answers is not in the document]
filariow: If we chose to use "FAILED" and "ACTIVE" it would also make sense to have them as constants in https://github.com/konflux-ci/multi-platform-controller/blob/6cb9b99665a2b14480dbe2f775aedb6cb4b82e5c/pkg/cloud/cloud.go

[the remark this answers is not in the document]
enkeefe00: You read my mind lol. I just implemented this.

pkg/cloud/cloud.go
```diff
@@ -7,7 +7,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const InstanceTag = "multi-platform-instance"
+const (
+	InstanceTag = "multi-platform-instance"
+	OKState     = "OK"
+	FailedState = "FAILED"
```
sadlerap, line 13: Might be good for a future PR, but generally if you have related values indicating some kind of state, the type of the values you're interested in should be wrapped in some kind of newtype pattern.  For instance, somewhere you'll have something like:
```go
type State string
const (
    OKState State     = "OK"
    FailedState State = "FAILED"
)
```

The benefit here is that it's clear that these values aren't arbitrary string, but rather can only have certain predefined values.

sadlerap (commented):
one small comment you can punt to a future PR if you wish, otherwise lgtm

pkg/cloud/cloud.go
enkeefe00: Cool, that's good information to know and am happy to implement it here. :smile:

sadlerap (approved):

mshaposhnik (approved):

PR #453, opened 2025-04-15
Make concurrent reconciliations configurable

mshaposhnik:
We have set 10 as an emergency action,  but now it turns that we're not always need that much (and even can be bad for some types of builds)

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/konflux-ci/multi-platform-controller/pkg/controller"
+	k8scontroller "sigs.k8s.io/controller-runtime/pkg/controller"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	//+kubebuilder:scaffold:imports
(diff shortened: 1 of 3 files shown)


cmd/controller/main.go
```diff
@@ -43,6 +45,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.IntVar(&concurrentReconciles, "concurrent-reconciles", 10, "The concurrency level for reconciling resources.")
```
filariow, line 48: should we default to a smaller value?

```suggestion
	flag.IntVar(&concurrentReconciles, "concurrent-reconciles", 1, "The concurrency level for reconciling resources.")
```

mshaposhnik: nope, 1 is too low. we will need to make patches if infra-depl for every installation.
so i would left it as 10 but reduce where it is needed (so far ocp-art only)

hugares: To keep same behavior as today, I would keep default to 10

filariow (approved):
/lgtm

enkeefe00 (approved):

PR #456, opened 2025-04-23
ci: upload cover.out to codecov

sadlerap:
Codecov is currently not doing much since we're not telling it where to find coverage data.  Since we're generating it, might as well use it.

### DIFF
--- a/.github/workflows/mpc-test-sealights.yaml
+++ b/.github/workflows/mpc-test-sealights.yaml
@@ -93,6 +93,8 @@ jobs:
         run: make test
       - name: Codecov
         uses: codecov/codecov-action@84508663e988701840491b86de86b666e8a86bed # v4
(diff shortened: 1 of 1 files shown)


sadlerap:
As an example, see [this warning](https://github.com/konflux-ci/multi-platform-controller/actions/runs/14590081021/job/40923122671?pr=452#step:16:52) from PR #452.

meyrevived:
Technically, it's good that we don't have warnings in our workflows.
But codecov has only been in our CI because we didn't have SeaLights. SeaLights has better granular analysis of code coverage and it is also pushed hard by The Management Above.
Perhaps remove it all together, as long as nothing is using it. Better make sure, though, by asking on the forum-konflux-developers and tagging Michael, Gal and Danny Baez (sp?)

PR #458, opened 2025-04-25
Fix host pool update tests

sadlerap:
Our host pool tests aren't testing what we think they're testing, so we could stand to improve them a bit.  The highlights:

- Assert that secrets are what we expect them to be, not that they're shaped like itself.
- Remove some tests that are actually race conditions in disguise (and fix the test to ideally not run into the race condition).
- Add a test to exercise another path through `UpdateHostPools`, so we should have more coverage.

### DIFF
--- a/pkg/reconciler/taskrun/hostupdate_test.go
+++ b/pkg/reconciler/taskrun/hostupdate_test.go
@@ -18,6 +18,10 @@
 package taskrun
 
 import (
+	"context"
+	"sync"
+	"time"
+
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -28,6 +32,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
 
 const testNamespace = "default"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -98,6 +101,32 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
 		}
 	})
 
+	It("should fail when the config doesn't exist", func(ctx SpecContext) {
```
meyrevived, line 104: Is this an edge-case scenario or something you're preparing towards implementing you solution to replace host-config.yaml?

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -155,47 +155,5 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
 			"user":        "root",
 			"platform":    "linux/ppc64le"},
 			"host.", true),
-		Entry("Negative test - bad address field", map[string]string{
```
meyrevived, line 158: There're also passing because I wrote them to pass. There's no real config data verification, but as soon as there is one, we can switch them back on to verify them.

meyrevived (changes requested):
Looks good, just have a couple of questions

pkg/reconciler/taskrun/hostupdate_test.go
sadlerap: I noticed we didn't have coverage on the first branch in the code, so I added a test to increase coverage.  I doubt it's actually something we're going to need to care about in production deployments.

meyrevived: Well, if static machines are making a comeback, we never know what other code might be brought out of the mothballs. If you can think of more scenarios test test this function with, please write them.

pkg/reconciler/taskrun/hostupdate_test.go
sadlerap: Ok, sitting down and thinking a bit more about these assertions.

The way I see it, there's a few ways we could do this validation in a way tests can catch, and all of them have problems:
1. We can run these tests against a cluster with tekton running, since both the CRDs and webhooks will return validation errors when we try to copy things into the taskrun we submit to the cluster.  I think trying to spin up any kind of cluster is going to be a non-starter for unit tests, though it might make sense for some higher-level tests.
2. We can call the [Validate] method on taskruns and do taskrun validation ourselves, rather than punting it to tekton's admission controllers.  However, it feels a little redundant when tekton's controllers are already doing these checks.  It also feels like we're testing tekton's validation capabilities, not our code.
3. We can validate things as we pull them out of the host config.  I think this is going to constantly be out-of-sync with whatever tekton does, and it's going to be annoying to maintain.  I'd rather not do this if we can help it.

Of the three, I dislike the second option the least.  If you want, I can add it in.  What do you think?

In any case, I think `UpdateHostPools` is either going to need to be fallible (returns some kind of error) or we need to log errors when the TaskRun creation fails.  We currently have no way of reporting that, and I think that's not right.

[Validate]: https://pkg.go.dev/github.com/tektoncd/pipeline@v1.0.0/pkg/apis/pipeline/v1#TaskRun.Validate

meyrevived: I don't see how TaskRun.Validate can help us fix the code to make these specific tests pass - you can validate _some_ field types,  but you can't validate a new host IP to be a valid and properly formatted IPv4 address.

The [planned acceptance tests](https://miro.com/app/board/uXjVIJXwS2g=/) are going to be run on a kind Konflux cluster , maybe that's how we  can implement the first suggestion?

The third suggestion implies host-config.yaml is here to stay, just when I started banking on it not doing so.

sadlerap: Yeah, that's why I'm not a fan of option 3.  I'm hoping to deprecate host-config.yaml and replace it with CRDs, but that's well outside the scope of this PR.  We might be able to punt some of these validations as field validations built into the CRD, but I'm not 100% sure about all of them.

meyrevived: Actually, for these simple fields, kyverno might be a good option. If we don't have a huge array to compare a match and remember matches we found because we need to make sure now two platform strings have been in the array before, and when it's a CRD floating around, Kyverno would not be a problem.

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -92,18 +101,67 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
...
+					}, obj.GetName())
+				},
+			}).Build()
+
+		log := logr.FromContextOrDiscard(ctx)
+		UpdateHostPools(testNamespace, k8sClient, &log)
+
+		list := v1.TaskRunList{}
+		Expect(k8sClient.List(ctx, &list)).To(Succeed())
+		Expect(list.Items).To(HaveLen(0))
```
filariow, line 127:
```suggestion
		// given the host config configmap doesn't exist
		k8sClient := fake.NewClientBuilder().
			WithScheme(scheme).
			WithInterceptorFuncs(interceptor.Funcs{
				Get: func(
					ctx context.Context,
					client client.WithWatch,
					key types.NamespacedName,
					obj client.Object,
					opts ...client.GetOption,
				) error {
					return errors.NewNotFound(schema.GroupResource{
						Group:    obj.GetObjectKind().GroupVersionKind().Group,
						Resource: "configmaps",
					}, obj.GetName())
				},
			}).Build()

		// when we run the UpdateHostPools is executed
		log := logr.FromContextOrDiscard(ctx)
		UpdateHostPools(testNamespace, k8sClient, &log)

		// then no Update TaskRuns are created
		list := v1.TaskRunList{}
		Expect(k8sClient.List(ctx, &list)).To(Succeed())
		Expect(list.Items).To(HaveLen(0))
```

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -92,18 +101,67 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
...
+					key types.NamespacedName,
+					obj client.Object,
+					opts ...client.GetOption,
+				) error {
+					return errors.NewNotFound(schema.GroupResource{
+						Group:    obj.GetObjectKind().GroupVersionKind().Group,
+						Resource: "configmaps",
+					}, obj.GetName())
+				},
+			}).Build()
```
filariow, line 120: We don't need this, as we are creating a new Client into which the ConfigMap have never been registered to

```suggestion
			Build()
```

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -92,18 +101,67 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
...
+
 	DescribeTable("Creating taskruns for updating static hosts in a pool",
 		func(ctx SpecContext, hostConfigData map[string]string, hostSuffix string, shouldFail bool) {
 
 			hostConfig.Data = testConfigDataFromTestData(hostConfigData, hostSuffix)
 
-			k8sClient := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(hostConfig).Build()
+			if !shouldFail {
+				waitGroup.Add(1)
+			}
```
filariow, line 137: I'd suggest to add some comments to clarify how `waitGroup` is used here

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -92,18 +101,67 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
...
+		Expect(list.Items).To(HaveLen(0))
+	})
+
 	DescribeTable("Creating taskruns for updating static hosts in a pool",
 		func(ctx SpecContext, hostConfigData map[string]string, hostSuffix string, shouldFail bool) {
 
 			hostConfig.Data = testConfigDataFromTestData(hostConfigData, hostSuffix)
 
-			k8sClient := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(hostConfig).Build()
+			if !shouldFail {
```
filariow, line 135: personally, I dislike having this kind of checks. I'd vote for splitting this test in two: failing ones, passing ones.

filariow (commented):
great work, some ideas

pkg/reconciler/taskrun/hostupdate_test.go
sadlerap: Fixed in c6ab255.

pkg/reconciler/taskrun/hostupdate_test.go
sadlerap: I've run with this idea in c6ab255, PTAL.

meyrevived:
@sadlerap do you have an idea why we're getting failures on [this test](https://github.com/konflux-ci/multi-platform-controller/pull/458/checks#step:15:288) suddenly? I've re-run the tests here and it fails on this spec again.

sadlerap:
It would appear the machine behind 150.239.19.36 is no longer responding to TCP connections on port 22.  Both tcpdump and wireshark seem to indicate that they don't even respond to tcp SYN packets - we're probably getting filtered.

We may need to find an IP address that responds to these connections for that test to function again, ideally on an IP address we control.

sadlerap:
Either way, that's outside the scope of this PR.

meyrevived:
It's actually a ppc machine and it's probably down because of the maintenance thing. I know the test says this is AWS, but Barak kept deleting the machine because he thought it was a pointless test but I wanted some kind of a canary back when we had connection problems with IBM.

pkg/reconciler/taskrun/hostupdate_test.go
```diff
@@ -92,110 +97,130 @@ var _ = Describe("HostUpdateTaskRunTest", func() {
...
+				"user":        "koko_hazamar",
+				"platform":    "linux/ppc64le"},
+				"dynamic.moshe-kipod-prod-1.", true),
+			Entry("bad key format", map[string]string{
+				"address":     "10.130.75.23",
+				"secret":      "internal-prod-ibm-ssh-key",
+				"concurrency": "1",
+				"user":        "root",
+				"platform":    "linux/ppc64le"},
+				"host.", true),
```
filariow, line 223: the `shouldFail` flag is no more used

```suggestion
			func(ctx SpecContext, hostConfigData map[string]string, hostSuffix string) {
				// In these tests, we have no way of synchronizing any spawned goroutines
				// with this thread, since we do no expect any to be spawned.  Instead, we
				// will expect the calls to list all taskruns to succeed multiple times with
				// some delays between checks. This is not an ideal check, but it works in
				// practice.  Having these checks helps prevent buggy tests and buggy
				// implementations of UpdateHostPools.

				// given: an invalid host config
				hostConfig.Data = testConfigDataFromTestData(hostConfigData, hostSuffix)

				k8sClient := fake.NewClientBuilder().
					WithScheme(scheme).
					WithRuntimeObjects(hostConfig).
					Build()

				// when: host pools are updated
				log := logr.FromContextOrDiscard(ctx)
				UpdateHostPools(testNamespace, k8sClient, &log)

				// test everything in TaskRun creation that is not part of the table testing
				Eventually(func(g Gomega) {
					createdList := v1.TaskRunList{}
					g.Expect(k8sClient.List(ctx, &createdList, client.InNamespace(testNamespace))).To(Succeed())
					g.Expect(createdList.Items).To(BeEmpty())
				}).
					MustPassRepeatedly(3).
					ProbeEvery(time.Second).
					Within(10 * time.Second).
					Should(Succeed())
			},
			Entry("empty data map", map[string]string{}, ""),
			Entry("dynamic host keys", map[string]string{
				"address":     "10.130.75.23",
				"secret":      "internal-moshe-kipod-ssh-key",
				"concurrency": "1",
				"user":        "koko_hazamar",
				"platform":    "linux/ppc64le"},
				"dynamic.moshe-kipod-prod-1."),
			Entry("bad key format", map[string]string{
				"address":     "10.130.75.23",
				"secret":      "internal-prod-ibm-ssh-key",
				"concurrency": "1",
				"user":        "root",
				"platform":    "linux/ppc64le"},
				"host."),
```

filariow (approved):
overall lgtm. Just a minor suggestion. thanks for looking into this!

pkg/reconciler/taskrun/hostupdate_test.go
sadlerap: fixed in da2309c

PR #460, opened 2025-05-05
KFLUXINFRA-949 - creating a unit test for otp.go -> ServeHTTP

meyrevived:
A unit test for storekey.ServeHTTP and otp.ServeHTTP.
Tests various happy paths for each, each for their own specific functionality, and some borderline or possibly problematic situations:

1. otp.ServeHTTP checks it can provision a one time password properly when given an ssh key, and checks that if globalMap does not have a matching otp value, it returns the correct error and log writes.
2. storekey.ServeHTTP checks it can store a valid ssh key in globalMap, that it can store borderline-but-valid ssh keys in globalMap.
3. storekey.ServeHTTP + otp.ServeHTTP - a third set of tests is an end-to-end test spec to make sure two ssh keys do not get the same OTP and that two different OTPs don't get the same ssh key in return.

### DIFF
--- a/cmd/otp/otp_servehttp_test.go
+++ b/cmd/otp/otp_servehttp_test.go
@@ -103,8 +103,8 @@ var _ = Describe("ServeHTTP handlers", Serial, func() {
 			})
 
 			DescribeTable("valid-but-borderline SSH keys - stores the SSH key in OTP map when given a valid key",
-				func(inputFn func() string, label string) {
-					input := inputFn()
+				func(input string, label string) {
+					//input := inputFn()
 					GinkgoWriter.Printf("Running test for: %s\n", label)
 					GinkgoWriter.Printf("Valid key value is: %s\n", input)
 
@@ -116,10 +116,10 @@ var _ = Describe("ServeHTTP handlers", Serial, func() {
 					Expect(logCapture.Contains("stored SSH key in OTP map")).To(BeTrue(), "Log should indicate success of SSH key placement in globalMap")
 				},
 
-				Entry("key with trailing whitespace", func() string { return rsaKey + "\n\t" }, "rsa key with whitespace at the end"),
-				Entry("key with comment", func() string { return rsaKey + " user@example.com" }, "rsa key with comment at the end"),
-				Entry("very long key", func() string { return rsaKey + rsaKey + rsaKey }, "a very long rsa key"),
-				Entry("ed25519 key", func() string { return edKey }, "an ed25519 key"),
+				Entry("key with trailing whitespace", rsaKey+"\n\t", "rsa key with whitespace at the end"),
+				Entry("key with comment", rsaKey+" user@example.com", "rsa key with comment at the end"),
+				Entry("very long key", rsaKey+rsaKey+rsaKey, "a very long rsa key"),
+				Entry("ed25519 key", edKey, "an ed25519 key"),
 			)
 		})
 
--- a/cmd/otp/otp_servehttp_test.go
+++ b/cmd/otp/otp_servehttp_test.go
@@ -55,17 +55,19 @@ var _ = Describe("ServeHTTP handlers", Serial, func() {
 		Context("with valid SSH key (otp)", func() {
 
 			It("serves a one-time password", func() {
-				// stores the valid ssh key in globalMap
+				By("Initializing a new store with a valid ssh key")
 				store := NewStoreKey(&logCapture.Logger)
 				storeReq := httptest.NewRequest("POST", "/store", strings.NewReader(rsaKey))
 				store.ServeHTTP(rr, storeReq)
 				storedKey := rr.Body.String()
 				GinkgoWriter.Printf("Stored key used for OTP: %q\n", storedKey)
 
+				By("Requiring an OTP for that ssh key")
 				req = httptest.NewRequest("POST", "/otp", strings.NewReader(storedKey))
 				rr = httptest.NewRecorder()
 				testOtp.ServeHTTP(rr, req)
 
+				By("Verifying that the OTP is returned successfully")
 				Expect(rr.Code).To(Equal(http.StatusOK))
 				Expect(rr.Body.String()).NotTo(BeEmpty())
 				Expect(logCapture.Contains("served one time password")).To(BeTrue(), "Log should indicate success of one time password provision")
--- a/cmd/otp/otp_servehttp_test.go
+++ b/cmd/otp/otp_servehttp_test.go
@@ -128,28 +128,6 @@ var _ = Describe("ServeHTTP handlers", Serial, func() {
 				Entry("ed25519 key", func() string { return edKey }, "an ed25519 key"),
 			)
 		})
-
-		// Skipping while KFLUXINFRA-1569 in SSH key validation is being fixed
-		PContext("with invalid ssh key (storekey)", func() {
-			badPrefixKey := "ssh-xyz123 AAAAB3NzaC1yc2EAAAADAQABAAABAQCe72TIPRwW/tDWPbNdJg3a6Rqy1/9kM002NLCx82fAN8GTvUj6VAD4Nl9om5BT7o0tfCcYgpDmTDrl7QPhBGd5ew0VKHO8o++SwhG6QI2mR+867qIdXP1B1ZdxO8eYndIn+ssOZcmbp4XxrI5/xWSNAU2XMckuSeUFZRTDNUVoKYmDMgnZL+BV6eQKO4jJmtuctDku4yb7YjcJYw7L+LOU7fBnsEpzvPJ/P9pGckpVI5nYIdQIUBxmVloa6BiKCGbu4yzPZ2zWISPODyvv6cmKk3s+YGild/TrVC+aRRN4TpiYq5NVT0lkkHpzhShVPukOR4xXoqZQLWYwAPLuUIBn"
-			DescribeTable("returns 500 for invalid SSH keys (storekey)",
-				func(input string, label string) {
-					GinkgoWriter.Printf("Running test for: %s\n", label)
-					GinkgoWriter.Printf("Invalid key value is: `%s`\n", input)
-
-					req := httptest.NewRequest("POST", "/store", strings.NewReader(input))
-
-					testStorekey.ServeHTTP(rr, req)
-					Expect(rr.Code).To(Equal(http.StatusInternalServerError))
-					Expect(logCapture.Contains("failed to read request body")).To(BeTrue(), "Log should indicate a problem with the request's body")
-				},
-
-				Entry("empty string", "", "An empty string"),
-				Entry("non-ssh string", "not-an-SSH-key", "A non-ssh key string"),
-				Entry("bad prefix", badPrefixKey, "An ssh key string with a badly formed prefix"),
-				Entry("missing payload", "ssh-rsa AAAAB3Nz", "An ssh key string with truncated content"),
-			)
-		})
 	})
 
 	Describe("End-to-end otp flow using ServeHTTP", func() {
--- a/cmd/otp/otp_servehttp_test.go
+++ b/cmd/otp/otp_servehttp_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"crypto"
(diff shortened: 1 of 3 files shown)


cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+				},
+
+				Entry("key with trailing whitespace", func() string { return rsaKey + "\n\t" }, "rsa key with whitespace at the end"),
+				Entry("key with comment", func() string { return rsaKey + " user@example.com" }, "rsa key with comment at the end"),
+				Entry("very long key", func() string { return rsaKey + rsaKey + rsaKey }, "a very long rsa key"),
+				Entry("ed25519 key", func() string { return edKey }, "an ed25519 key"),
+			)
+		})
+
+		// Skipping while KFLUXINFRA-1569 in SSH key validation is being fixed
```
ifireball, line 126: I don't think key validation belongs in the OTP service codebase, so I would prefer not to keep this test, and also possibly simplify other test code by using random strings rather then valid keys.

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+				testStorekey.ServeHTTP(rr, req)
+
+				Expect(rr.Code).To(Equal(http.StatusOK))
+				Expect(rr.Body.String()).NotTo(BeEmpty())
+				Expect(logCapture.Contains("stored SSH key in OTP map")).To(BeTrue(), "Log should indicate success of SSH key placement in globalMap")
+			})
+
+			DescribeTable("valid-but-borderline SSH keys - stores the SSH key in OTP map when given a valid key",
+				func(inputFn func() string, label string) {
+					input := inputFn()
```
sadlerap, line 107: Why do we take the input as a function returning string instead of a string?  This seems like unnecessary indirection to me, and it might let us compact construction of our entries below if we don't do this.

sadlerap (approved):
I have one question, but it's more confusion on my part more than anything.  Tests look great!

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+		edKey, err = generateValidSSHKey("ed25519")
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	Describe("Testing otp.ServeHTTP", func() {
+		var testOtp *otp
+
+		JustBeforeEach(func() {
+			testOtp = NewOtp(&logCapture.Logger)
+		})
```
filariow, line 53: why do we need this in a `JustBeforeEach`? Isn't `BeforeEach` enough?

```suggestion
		BeforeEach(func() {
			testOtp = NewOtp(&logCapture.Logger)
		})
```

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+		Context("with valid SSH key (otp)", func() {
+
+			It("serves a one-time password when given a valid key", func() {
```
filariow, line 57: It seems to me this is redundant.

```suggestion
		Context("with valid SSH key (otp)", func() {

			It("serves a one-time password", func() {
```

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+	})
+
+	Describe("Testing otp.ServeHTTP", func() {
+		var testOtp *otp
+
+		JustBeforeEach(func() {
+			testOtp = NewOtp(&logCapture.Logger)
+		})
+
+		Context("with valid SSH key (otp)", func() {
```
filariow, line 55: I'd suggest breaking this in two Ordered `It` containers. Something like 

```go
		Context("with valid SSH key (otp)", Ordered, func() {
                        var storedKey string

			It("serves key stored for one-time password", func() {
			....
			It("serves one-time password for provided key", func() {
			....
```

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+				Expect(logCapture.Contains("no OTP found for provided SSH key")).To(BeTrue(),
+					"Log should indicate the SSH key wasn't found in the store")
+			})
+		})
+	})
+
+	Describe("Testing storekey.ServeHTTP", func() {
+		var testStorekey *storekey
+
+		JustBeforeEach(func() {
```
filariow, line 90: ditto

```suggestion
		BeforeEach(func() {
```

cmd/otp/otp_servehttp_test.go
meyrevived: Because for some *content censured* reason the damn string concatination didn't work! At first I thought it was just something wrong with GinkgoWriter, but then I put a debug `fmt.PrintF(string(body))` in otp.go right after the body is extracted from the request and I saw that concatinated strings weren't making it into ServeHTTP itself either.
This *content censured* is the only things that *content censured* worked.

cmd/otp/otp_servehttp_test.go
filariow: I agree, this seems a new feature to me. Let's discuss on the JIRA ticket and -in case- open a dedicated PR for that.

cmd/otp/otp_servehttp_test.go
```diff
@@ -0,0 +1,199 @@
...
+			)
+		})
+	})
+})
+
+// Generates a real SSH public key string, with option to change the encryption algorithm. Can be broadened to more
+// algorithms such as diffie-hellman, if we ever choose to test it.
+func generateValidSSHKey(keyType string) (string, error) {
+	switch keyType {
+	case "rsa":
```
filariow, line 154: I'd suggest to have a some constants here

```go
const (
	rsaKeyType	string = "rsa"
	ed25519KeyType	string = "ed25519"
)
```

```suggestion
	case rsaKeyType:
```

filariow (changes requested):
mostly lgtm, some feedback

cmd/otp/otp_servehttp_test.go
filariow: this works on my side

```diff
                        DescribeTable("valid-but-borderline SSH keys - stores the SSH key in OTP map when given a valid key",
-                               func(inputFn func() string, label string) {
-                                       input := inputFn()
+                               func(input string, label string) {
                                        GinkgoWriter.Printf("Running test for: %s\n", label)
                                        GinkgoWriter.Printf("Valid key value is: %s\n", input)

@@ -116,10 +114,10 @@ var _ = Describe("ServeHTTP handlers", Serial, func() {
                                        Expect(logCapture.Contains("stored SSH key in OTP map")).To(BeTrue(), "Log should indicate success of SSH key placement in globalMap")
                                },

-                               Entry("key with trailing whitespace", func() string { return rsaKey + "\n\t" }, "rsa key with whitespace at the end"),
-                               Entry("key with comment", func() string { return rsaKey + " user@example.com" }, "rsa key with comment at the end"),
-                               Entry("very long key", func() string { return rsaKey + rsaKey + rsaKey }, "a very long rsa key"),
-                               Entry("ed25519 key", func() string { return edKey }, "an ed25519 key"),
+                               Entry("key with trailing whitespace", rsaKey+"\n\t", "rsa key with whitespace at the end"),
+                               Entry("key with comment", rsaKey+" user@example.com", "rsa key with comment at the end"),
+                               Entry("very long key", rsaKey+rsaKey+rsaKey, "a very long rsa key"),
+                               Entry("ed25519 key", edKey, "an ed25519 key"),
```

meyrevived: Maybe it's an IDE issue, because this is what I was doing initially and it did not work. In any case, I didn't want to risk it.

meyrevived: Nope, just tried reverting to the way it was before - as you say works for you - and put back the request body print line and I don't see the concatenated strings:
image

cmd/otp/otp_servehttp_test.go
meyrevived: We're already testing storedKey functionality with a valid SSH key in line 94. I added a storekey implementation here because otp.ServeHTTP needs to have the ssh key in globalMap in order to look for it, or I get a false negative because I'd get to this situation:
`if !loaded {
		s.logger.Error(err, "no OTP found for provided SSH key", "address", request.RemoteAddr)
		writer.WriteHeader(400)
}`

cmd/otp/otp_servehttp_test.go
meyrevived: See my response to the previous comment

cmd/otp/otp_servehttp_test.go
meyrevived: Not sure what the benefits of it are. 
It's not like we're going to change the value of these strings, and we want to change the code in only one place and have it effect multiple places. 
Also, these are very specific algorithm names - they're not a timeout value or a value that might change in the future without also changing it's name. 
If we decide to mostly test ed25519 instead of rsa and switch ed25519 with diffie-hellman, we'll also have to change  <name>KeyType.

cmd/otp/otp_servehttp_test.go
meyrevived: Now, with the (ugly, I know) ` func() string`  thing, I do see the concatenated string:
image

cmd/otp/otp_servehttp_test.go
filariow: I think we should we also check that two subsequent invocation of the `/otp` endpoint returns different values

filariow (changes requested):

cmd/otp/otp_servehttp_test.go
meyrevived: I thought, style-wise, that it is another way to compartmentalize the two test sessions for otp.ServeHTTP and storekey.HTTP. Also, saving memory and processing power, even in the slightest way. I spent two years developing a C++ real-time product, every bit mattered, and I guess the habit stuck.

cmd/otp/otp_servehttp_test.go
meyrevived: I'll +1 you on that - ginkgo has [a tool for it](https://onsi.github.io/ginkgo/#documenting-complex-specs-by)

See the [latest commit](https://github.com/konflux-ci/multi-platform-controller/pull/460/commits/6412436936c7cbc67dcfe0b3d0a57caa9f3d0eb2)

cmd/otp/otp_servehttp_test.go
meyrevived: OK. Not a hill I'm willing to die on. See latest commit.

cmd/otp/otp_servehttp_test.go
meyrevived: There's already [a unit test](https://github.com/meyrevived/multi-platform-controller/blob/5daeeccfcdf21bac56891df4c001fd42f8d26f84/cmd/otp/otp_test.go#L18) for GenerateRandomString, to make sure it doesn't generate the same random string.

cmd/otp/otp_servehttp_test.go
filariow: great!

cmd/otp/otp_servehttp_test.go
filariow: @meyrevived I think we can remove this for now. Let's introduce it in future if needed. We can take note of this code in the JIRA ticket.

Also, if request content is invalid, we should return `400` as the error is on the user side.

cmd/otp/otp_servehttp_test.go
filariow: the test for GenerateRandomString is testing the util function used by the HTTP handler. is something changes in the handler or if it introduces a bug we can not tell with just that test. It's not critical, but it may worth to have a small test here for that.

meyrevived: See latest commit for a test implemented about this

cmd/otp/otp_servehttp_test.go
meyrevived: I missed this somehow, applied changes in the latest commit

filariow (approved):
/lgtm

PR #465, opened 2025-05-13
tests: fix ginkgo test coverage

sadlerap:
Newer versions (> v2.23 as far as I can tell) of ginkgo expect all command-line arguments before packages.  Fix invocations of ginkgo to abide by these rules.


meyrevived (approved):
/lgtm

PR #466, opened 2025-05-14
Kfluxinfra-1587 work

meyrevived:
A unit test for ec2_helpers.go -> configureInstance.
Breaks the various fields configureInstance loads into ec2.RunInstancesInput into families of fields and tests each in different Contexts. 
Sad path tests are limited in the scope of the actual answer they give. Since configureInstance doesn't return an error, and in the scope of a unit test we're not going to attempt to actually launch an EC2 machine with the generated ec2.RunInstancesInput instance, so we won't know for sure what breaks if we give configureInstance an empty AMI field, etc.

### DIFF
--- a/pkg/aws/ec2_helpers_test.go
+++ b/pkg/aws/ec2_helpers_test.go
@@ -230,16 +230,13 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 			DescribeTable("with various subnet and security group settings",
 				// This DescribeTable takes an ecConfig-mutating function and a function that verifies the new content
 				// of the ecConfig after going through configureInstance
-				func(setupConfig func(conf *AWSEc2DynamicConfig), verify func(input *ec2.RunInstancesInput)) {
-					currentConfig := ecConfig
-					setupConfig(&currentConfig)
+				func(setupConfig func() AWSEc2DynamicConfig, verify func(input *ec2.RunInstancesInput)) {
+					currentConfig := setupConfig()
 					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
 					verify(runInput)
 				},
 				Entry("using baseline SubnetId and SecurityGroupId",
-					func(conf *AWSEc2DynamicConfig) {
-						// ecConfig is as a valid default EC2 Config
-					},
+					newDefaultValidEC2ConfigForInstance(),
 					func(input *ec2.RunInstancesInput) {
 						Expect(input.SubnetId).To(Equal(&ecConfig.SubnetId))
 						Expect(input.SecurityGroupIds).To(ContainElement(ecConfig.SecurityGroupId))
@@ -247,9 +244,11 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 					},
 				),
 				Entry("with only SecurityGroup name (clearing SG ID from baseline)",
-					func(conf *AWSEc2DynamicConfig) {
+					func() AWSEc2DynamicConfig {
+						conf := newDefaultValidEC2ConfigForInstance()
 						conf.SecurityGroup = "sg-name-example"
 						conf.SecurityGroupId = ""
+						return conf
 					},
 					func(input *ec2.RunInstancesInput) {
 						Expect(input.SecurityGroups).To(ContainElement("sg-name-example"))
@@ -257,10 +256,12 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 					},
 				),
 				Entry("with no network settings (clearing baseline)",
-					func(conf *AWSEc2DynamicConfig) {
+					func() AWSEc2DynamicConfig {
+						conf := newDefaultValidEC2ConfigForInstance()
 						conf.SubnetId = ""
 						conf.SecurityGroupId = ""
 						conf.SecurityGroup = ""
+						return conf
 					},
 					func(input *ec2.RunInstancesInput) {
 						Expect(input.SubnetId).To(BeNil())
@@ -275,16 +276,13 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 			DescribeTable("with various IAM profile settings",
 				// This DescribeTable also takes an ecConfig-mutating function and a function that verifies the new
 				//content of the ecConfig after going through configureInstance
-				func(setupConfig func(conf *AWSEc2DynamicConfig), verify func(input *ec2.RunInstancesInput)) {
-					currentConfig := ecConfig
-					setupConfig(&currentConfig)
+				func(setupConfig func() AWSEc2DynamicConfig, verify func(input *ec2.RunInstancesInput)) {
+					currentConfig := setupConfig()
 					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
 					verify(runInput)
 				},
 				Entry("using baseline InstanceProfileArn",
-					func(conf *AWSEc2DynamicConfig) {
-						// ecConfig ia as it's created in BeforeEach.
-					},
+					newDefaultValidEC2ConfigForInstance(),
 					func(input *ec2.RunInstancesInput) {
 						Expect(input.IamInstanceProfile).NotTo(BeNil())
 						Expect(input.IamInstanceProfile.Arn).To(Equal(&ecConfig.InstanceProfileArn))
@@ -292,9 +290,11 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 					},
 				),
 				Entry("with only InstanceProfileName (clearing ARN from baseline)",
-					func(conf *AWSEc2DynamicConfig) {
+					func() AWSEc2DynamicConfig {
+						conf := newDefaultValidEC2ConfigForInstance()
 						conf.InstanceProfileName = "profile-name-example"
 						conf.InstanceProfileArn = ""
+						return conf
 					},
 					func(input *ec2.RunInstancesInput) {
 						Expect(input.IamInstanceProfile).NotTo(BeNil())
@@ -304,9 +304,11 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 				),
 				// TODO: scenarios like missing AMI fields need to be tested properly. Not in the scope of this humble unit test
 				Entry("with no IAM profile settings (clearing baseline)",
-					func(conf *AWSEc2DynamicConfig) {
+					func() AWSEc2DynamicConfig {
+						conf := newDefaultValidEC2ConfigForInstance()
 						conf.InstanceProfileName = ""
 						conf.InstanceProfileArn = ""
+						return conf
 					},
 					func(input *ec2.RunInstancesInput) { Expect(input.IamInstanceProfile).To(BeNil()) },
 				),
--- a/pkg/aws/ec2_helpers_test.go
+++ b/pkg/aws/ec2_helpers_test.go
@@ -179,8 +179,12 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 			Entry("Negative test - not an IP address", "Not an IP address", true),
 		)
 	})
-
-	Describe("configureInstance", func() {
+	// A unit test for configureInstance. Bypasses CreateEc2CloudConfig's logic to isolate cases where its logic may be
+	// faulty, hence the helper function newDefaultValidEC2ConfigForInstance which provides a neutral all-good
+	// AWSEc2DynamicConfig with baseline and passing generic configurations.
+	// Breaks the various configuration fields to families of configuration topics, each tested in it own Context. Some
+	// have happy test paths, some have sad ones.
+	Describe("Testing configureInstance", func() {
 		var (
 			ecConfig       AWSEc2DynamicConfig
 			taskRunName    string
@@ -288,6 +292,7 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 						Expect(input.IamInstanceProfile.Arn).To(BeNil())
 					},
 				),
+				// TODO: scenarios like missing AMI fields need to be tested properly. Not in the scope of this humble unit test
 				Entry("with no IAM profile settings (clearing baseline)",
 					func(conf *AWSEc2DynamicConfig) {
 						conf.InstanceProfileName = ""
--- a/pkg/aws/ec2_helpers_test.go
+++ b/pkg/aws/ec2_helpers_test.go
@@ -238,7 +238,7 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
 				},
 				Entry("using baseline SubnetId and SecurityGroupId",
 					func(conf *AWSEc2DynamicConfig) {
-						// ecConfig is as it's created in BeforeEach.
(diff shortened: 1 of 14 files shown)


sadlerap (commented):
lgtm so far, one small comment

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,248 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+				Expect(runInput.KeyName).To(PointTo(Equal(ecConfig.KeyName)))
+				Expect(runInput.ImageId).To(PointTo(Equal(ecConfig.Ami)))
+				Expect(runInput.InstanceType).To(Equal(types.InstanceType(ecConfig.InstanceType)))
+			})
+
+			It("should set MinCount=1, MaxCount=1, EbsOptimized=true, and ShutdownBehavior=terminate", func() {
+				runInput := ecConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+				Expect(runInput.MinCount).To(PointTo(Equal(int32(1))))
+				Expect(runInput.MaxCount).To(PointTo(Equal(int32(1))))
+				Expect(runInput.EbsOptimized).To(PointTo(BeTrue()))
```
filariow, line 214: it seems to me we are hardcoding these values in two places: here and in the code. I'd suggest to hardcode them just in one place: the code.

Here it would also be nice to test some properties of the hardcoded values, like if `Max>=Min>=0`.

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,248 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+				// of the ecConfig after going through configureInstance
+				func(setupConfig func(conf *AWSEc2DynamicConfig), verify func(input *ec2.RunInstancesInput)) {
+					currentConfig := ecConfig
+					setupConfig(&currentConfig)
+					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+					verify(runInput)
+				},
+				Entry("using baseline SubnetId and SecurityGroupId",
+					func(conf *AWSEc2DynamicConfig) {
+						// ecConfig is as it's created in BeforeEach.
```
filariow, line 231: for clarity of the test, I'd suggest to state clearly what we want it to be as per this test. In this case we want it to be a `valid default EC2 Config`.

```suggestion
						// ecConfig is as a valid default EC2 Config
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,248 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+				Expect(runInput.EbsOptimized).To(PointTo(BeTrue()))
+				Expect(runInput.InstanceInitiatedShutdownBehavior).To(Equal(types.ShutdownBehaviorTerminate))
+			})
+		})
+
+		Context("when configuring networking", func() {
+			DescribeTable("with various subnet and security group settings",
+				// This DescribeTable takes an ecConfig-mutating function and a function that verifies the new content
+				// of the ecConfig after going through configureInstance
+				func(setupConfig func(conf *AWSEc2DynamicConfig), verify func(input *ec2.RunInstancesInput)) {
```
filariow, line 223: Proposal: For making these tests easier to understand, would it be an idea to have the config created by this function instead of changed? 

```suggestion
				func(setupConfig func() AWSEc2DynamicConfig, verify func(input *ec2.RunInstancesInput)) {
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,248 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+					setupConfig(&currentConfig)
+					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+					verify(runInput)
+				},
+				Entry("using baseline SubnetId and SecurityGroupId",
+					func(conf *AWSEc2DynamicConfig) {
+						// ecConfig is as it's created in BeforeEach.
+					},
+					func(input *ec2.RunInstancesInput) {
+						Expect(input.SubnetId).To(PointTo(Equal(ecConfig.SubnetId)))
```
filariow, line 234: AFAYK, if there's a way of removing most of these `PointTo`?

filariow (changes requested):
great job! some feedback

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,248 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+					if iops == nil {
+						Expect(ebsDevice.Iops).To(BeNil())
+					} else {
+						Expect(ebsDevice.Iops).To(PointTo(Equal(*iops)))
+					}
+					if throughput == nil {
+						Expect(ebsDevice.Throughput).To(BeNil())
+					} else {
+						Expect(ebsDevice.Throughput).To(PointTo(Equal(*throughput)))
+					}
```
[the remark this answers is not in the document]
meyrevived: Cool, I "yes, and"-ed it, see latest commit.

pkg/aws/ec2_helpers_test.go
meyrevived: I went over these following Andy's suggestion, and found only one place where we can replace it with something that's different and 100% does the same job.
I looked into Gomega's  `BeComparableTo` but it could only replace some fields that aren't handled by aws.String() or aws.Int32() (which create a brand new pointer). Look at the commit I put in following Andy's suggestion, it is the most I saw I could do.

pkg/aws/ec2_helpers_test.go
meyrevived: You mean change  `newDefaultValidEC2ConfigForInstance()`  so that it would get either a full set of fields to create a new AWSEc2DynamicConfig with, or stay the way it is (with an AWSEc2DynamicConfig that has default values) but take as input some sort of data structure with the names of the fields that needs to be updated and the values they need to have updated?

It doesn't look more readable or more economic in IO actions than what is currently there.
Is that what you meant?

pkg/aws/ec2_helpers_test.go
meyrevived: Do you want something like 

`const (
    DefaultMinInstanceCount int32 = 1
    DefaultMaxInstanceCount int32 = 1
    DefaultEbsOptimized     bool  = true
)`

Added at the top of ec2_helpers.go and used later in the tests? If that's the case, is it because are truly fundamental constants that might be referenced elsewhere or could be configurable in the future (even if not now)? 
If yes, then defining them as const like that  is cleaner. 
If no, and they're just "what AWS needs for this specific call" and unlikely to change or be referenced elsewhere, I don't know the difference in readability between additional 5 lines of code defining them as `const` or `var` versus three values that are repeated in only two times in the code. 

As for `Max>=Min>=0`, see latest commit.

enkeefe00 (approved):

pkg/aws/ec2_helpers_test.go
sadlerap: This one could be taken out:
```suggestion
						Expect(input.SubnetId).To(Equal(&ecConfig.SubnetId))
```

gomega's `Equal` matcher doesn't care if pointers are equal, it looks at the underlying object and uses `reflect.DeepEqual` to test for equality.  If the intention is to test for pointer equality, you may want to use [BeIdenticalTo](https://onsi.github.io/gomega/#beidenticaltoexpected-any) instead.

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+					verify(runInput)
+				},
+				Entry("using baseline InstanceProfileArn",
+					func(conf *AWSEc2DynamicConfig) {
+						// ecConfig ia as it's created in BeforeEach.
+					},
+					func(input *ec2.RunInstancesInput) {
+						Expect(input.IamInstanceProfile).NotTo(BeNil())
+						Expect(input.IamInstanceProfile.Arn).To(PointTo(Equal(ecConfig.InstanceProfileArn)))
```
sadlerap, line 290:
```suggestion
						Expect(input.IamInstanceProfile.Arn).To(Equal(&ecConfig.InstanceProfileArn))
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+		})
+
+		Context("when configuring tags", func() {
+			It("should include default and Name tags correctly with baseline config", func() {
+				runInput := ecConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+				Expect(runInput.TagSpecifications).To(HaveLen(1))
+				Expect(runInput.TagSpecifications[0].ResourceType).To(Equal(types.ResourceTypeInstance))
+				tags := runInput.TagSpecifications[0].Tags
+				Expect(tags).To(ContainElement(SatisfyAll(HaveField("Key", PointTo(Equal(MultiPlatformManaged))), HaveField("Value", PointTo(Equal("true"))))))
+				Expect(tags).To(ContainElement(SatisfyAll(HaveField("Key", PointTo(Equal(cloud.InstanceTag))), HaveField("Value", PointTo(Equal(instanceTag))))))
```
sadlerap, line 342:
```suggestion
				Expect(tags).To(ContainElement(SatisfyAll(HaveField("Key", PointTo(Equal(cloud.InstanceTag))), HaveField("Value", Equal(&instanceTag)))))
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+
+		Context("when configuring EBS block device mappings", func() {
+			It("should set EBS properties correctly based on baseline config", func() {
+				runInput := ecConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+				Expect(runInput.BlockDeviceMappings).To(HaveLen(1))
+				bdMapping := runInput.BlockDeviceMappings[0]
+				Expect(bdMapping.DeviceName).To(PointTo(Equal("/dev/sda1")))
+				Expect(bdMapping.Ebs).NotTo(BeNil())
+				Expect(bdMapping.Ebs.DeleteOnTermination).To(PointTo(BeTrue()))
+				Expect(bdMapping.Ebs.VolumeSize).To(PointTo(Equal(ecConfig.Disk)))
```
sadlerap, line 363:
```suggestion
				Expect(bdMapping.Ebs.VolumeSize).To(Equal(&ecConfig.Disk))
```

sadlerap (commented):
A few more places we can remove `PointTo` calls that actually save some space.  There are more places we could take it out, but the alternative ends up being longer than just calling `PointTo`.

With these changes, lgtm.

pkg/aws/ec2_helpers_test.go
filariow: no, I mean change the `setupConfig func` such that it doesn't change a structure -that in our case is initialized in `BeforeEach`- but it creates it.

Something like this:

```golang
				func(setupConfig func(conf *AWSEc2DynamicConfig), verify func(input *ec2.RunInstancesInput)) {
					currentConfig := ecConfig
					setupConfig(&currentConfig)
					runInput := currentConfig.configureInstance(taskRunName, instanceTag, additionalTags)
					verify(runInput)
				},
				Entry("using baseline SubnetId and SecurityGroupId",
					newDefaultValidEC2ConfigForInstance,
					func(input *ec2.RunInstancesInput) {
						Expect(input.SubnetId).To(PointTo(Equal(ecConfig.SubnetId)))
						Expect(input.SecurityGroupIds).To(ContainElement(ecConfig.SecurityGroupId))
						Expect(input.SecurityGroups).To(BeNil())
					},
				),
				Entry("with only SecurityGroup name (clearing SG ID from baseline)",
					func() AWSEc2DynamicConfig {
						conf := newDefaultValidEC2ConfigForInstance()
						conf.SecurityGroup = "sg-name-example"
						conf.SecurityGroupId = ""
						return conf
					},
					func(input *ec2.RunInstancesInput) {
						Expect(input.SecurityGroups).To(ContainElement("sg-name-example"))
						Expect(input.SecurityGroupIds).To(BeNil())
					},
				),
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+					Expect(runInput.MinCount).NotTo(BeNil())
+					Expect(runInput.MaxCount).NotTo(BeNil())
+					Expect(*runInput.MinCount).To(BeNumerically(">", 0),
+						"MinCount should be positive")
+					Expect(*runInput.MaxCount).To(BeNumerically(">=", *runInput.MinCount),
+						"MaxCount should be greater than or equal to MinCount")
+				})
+		})
+
+		Context("when configuring networking", func() {
```
filariow, line 229:
```suggestion
		When("configuring networking", func() {
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+					func(conf *AWSEc2DynamicConfig) {
+						conf.InstanceProfileName = ""
+						conf.InstanceProfileArn = ""
+					},
+					func(input *ec2.RunInstancesInput) { Expect(input.IamInstanceProfile).To(BeNil()) },
+				),
+			)
+		})
+
+		Context("when configuring spot instances", func() {
```
filariow, line 316:
```suggestion
		When("configuring spot instances", func() {
```

pkg/aws/ec2_helpers_test.go
```diff
@@ -175,5 +179,250 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+			It("should include default and additional tags when provided", func() {
+				additionalTags["CostCenter"] = "AlphaTeam"
+				runInput := ecConfig.configureInstance(taskRunName, instanceTag, additionalTags)
+				tags := runInput.TagSpecifications[0].Tags
+				Expect(tags).To(HaveLen(3 + len(additionalTags)))
+				Expect(tags).To(ContainElement(SatisfyAll(HaveField("Key", PointTo(Equal("CostCenter"))), HaveField("Value", PointTo(Equal("AlphaTeam"))))))
+			})
+		})
+
+		Context("when configuring EBS block device mappings", func() {
```
filariow, line 355:
```suggestion
		When("configuring EBS block device mappings", func() {
```

filariow (changes requested):
consistently replace `Context("when ` with `When("`

ifireball:
@meyrevived A little off-topic request please. You wrote a pretty good PR description, but the PR title is pretty much devoid of context. Could you please, in the future, include a bit more context in the PR title, so that when I get notifications about it in my inbox I know quickly what its about. Thanks!

PR #467, opened 2025-05-19
make: add lint and lint-go target

sadlerap:
We need a self-contained method of running golangci-lint as a make recipe.  In order to avoid installing via `go install` (a command for which we may not have sufficient privileges) and to avoid introducing it as a dependency in MPC, let's pull it in its own modfile and run it via `go run`.


meyrevived (approved):
/lgtm

PR #471, opened 2025-05-29
KFLUXINFRA-1588 - unit test for cloud.go -> ValidateTaskRunID

meyrevived:
A unit test for ValidateTaskRunID. For now only tests input that should not return an error from ValidateTaskRunID
and input that should return an error code, according to ValidateTaskRunID's current logic.
More testing scenarios are available, but they're currently exposing ValidateTaskRunID's bugs so are on hiatus and depend on [KFLUXINFRA-1697](https://issues.redhat.com/browse/KFLUXINFRA-1697) being fixed.

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -1,0 +1,108 @@
+package cloud
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("ValidateTaskRunID", func() {
+
+	// Testing everything that should pass, according to ValidateTaskRunID's current code
+	When("the TaskRunID is in the standard 'namespace:name' format", func() {
+		It("should return no error", func() {
+			input := "my-namespace:my-taskrun-123"
+			err := ValidateTaskRunID(input)
+			GinkgoWriter.Printf("Input: '%s' -> Expected: No Error, Got: %v\n", input, err)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+	// ValidateTaskRunID currently passes these, but shouldn't
+	When("the TaskRunID is structurally valid, but semantically wrong (empty parts)", func() {
+		DescribeTable("TaskRunID has one colon, but its actual parts are empty",
+			func(input string, description string) {
+				GinkgoWriter.Printf("Testing empty parts: Input: '%s' (Description: %s)\n", input, description)
+				err := ValidateTaskRunID(input)
+				GinkgoWriter.Printf("Input: '%s' -> Expected: No Error, Got: %v\n", input, err)
+				Expect(err).NotTo(HaveOccurred())
+			},
+			Entry("empty namespace part", ":a-taskrun-name", "e.g., ':taskrun-name'"),
+			Entry("empty name part", "a-namespace:", "e.g., 'namespace:'"),
+			Entry("just a single colon", ":", "e.g., ':'"),
+		)
+	})
+
+	When("the TaskRunID is structurally valid, but would be rejected by cloud provider' tagging rules", func() {
+		DescribeTable("TaskRunID has one colon and breaks AWS or IBM Cloud's tag string rules",
+			func(input string, cloudRejectionReason string) {
+				GinkgoWriter.Printf(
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,47 @@
...
+// the bug is tracked on Jira (ticket number inside the test spec's body).
+var _ = Describe("ValidateTaskRunID", func() {
+
+	// Testing everything that should pass, according to ValidateTaskRunID's current code
+	When("the TaskRunID is in the standard 'namespace:name' format", func() {
+		It("should return no error", func() {
+			input := "my-namespace:my-taskrun-123"
+			err := ValidateTaskRunID(input)
+			GinkgoWriter.Printf("Input: '%s' -> Expected: No Error, Got: %v\n", input, err)
+			Expect(err).Should(BeNil())
```
filariow, line 21: I think we can simplify this block a bit

```suggestion
			Expect(ValidateTaskRunID(input)).To(Succeed())
```

pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,47 @@
...
+		})
+	})
+
+	// When KFLUXINFRA-1697 is fixed, there will be more test scenarios here
+
```
filariow, line 26: As a general advice, I suggest to link PRs and code in JIRA issues rather than the adding comments in the code pointing to JIRA issue. If we decide not to implement a JIRA issue, we'll need to come here and clean the code too.

```suggestion
```

pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,47 @@
...
+		DescribeTable("it should return a formatted error",
+			func(input string, descriptionOfInvalidity string) {
+				GinkgoWriter.Printf("Testing structurally invalid scenario: Input: '%s' (Reason: %s)\n",
+					input, descriptionOfInvalidity)
+				err := ValidateTaskRunID(input)
+				Expect(err).Should(HaveOccurred())
+
+				expectedErrorMessage := fmt.Sprintf(
+					"'%s' does not follow the correct format: '<TaskRun Namespace>:<TaskRun Name>'", input)
+				Expect(err.Error()).To(Equal(expectedErrorMessage))
```
filariow, line 38: I'd suggest to simplify this as follow

```suggestion
				expectedErr := fmt.Errorf(
					"'%s' does not follow the correct format: '<TaskRun Namespace>:<TaskRun Name>'", input)
				Expect(ValidateTaskRunID(input)).To(MatchError(expectedErr), descriptionOfInvalidity)
```

filariow (changes requested):
well done, mostly lgtm. Some minor feedback to simplify code and maintenance

mshaposhnik:
What is the official way to run unit tests all-at-once  in the MPC repo? 
As a developer how can i locally make sure i didn't broke something?
`make tests` usually a standart command for that, but unfortunately it isn't working in MPC.

meyrevived:
@mshaposhnik  I just tried running `make test` from the terminal in Goland, VSCode and just my normal Fedora terminal and all three ways worked when I ran it from the root MPC repo directory.
What errors do you get when you try to run `make test`?
VSCode has the testing plugin enabled whenever the repo you have has tests, you can use it to run tests, and Goland has a [Ginkgo plugin](https://plugins.jetbrains.com/plugin/17554-ginkgo) you can run either the whole repo's tests or individual suites/specs through.

pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,40 @@
...
+// More testing scenarios are available, but they're currently exposing ValidateTaskRunID's bugs so are on hiatus and
+// the bug is tracked on Jira (ticket number inside the test spec's body).
+var _ = Describe("ValidateTaskRunID", func() {
+
+	// Testing everything that should pass, according to ValidateTaskRunID's current code
+	When("the TaskRunID is in the standard 'namespace:name' format", func() {
+		It("should return no error", func() {
+			input := "my-namespace:my-taskrun-123"
+			err := ValidateTaskRunID(input)
+			GinkgoWriter.Printf("Input: '%s' -> Expected: No Error, Got: %v\n", input, err)
```
filariow, line 20:
```suggestion
```

PR #491, opened 2025-07-04
add unit tests for taskrun.go's extractPlatform

filariow:
Assisted-by: Cursor
Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -166,7 +166,7 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 				},
 			}
 
-			Expect(extractPlatform(tr)).Error().To(MatchError("failed to determine platform"))
+			Expect(extractPlatform(tr)).Error().To(MatchError(errFailedToDeterminePlatform))
 		})
 
 		It("should return error when TaskRun has no parameters", func() {
@@ -177,7 +177,7 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 				},
 			}
 
-			Expect(extractPlatform(tr)).Error().To(MatchError("failed to determine platform"))
+			Expect(extractPlatform(tr)).Error().To(MatchError(errFailedToDeterminePlatform))
 		})
 
 		It("should return error when PlatformParam parameter is missing", func() {
@@ -190,7 +190,7 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 				},
 			}
 
-			Expect(extractPlatform(tr)).Error().To(MatchError("failed to determine platform"))
+			Expect(extractPlatform(tr)).Error().To(MatchError(errFailedToDeterminePlatform))
 		})
 	})
 
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -147,7 +147,6 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 		})
 
 		It("should return error when TaskRun has nil parameters", func() {
-			// Test case 9: TaskRun with nil parameters
 			tr := &pipelinev1.TaskRun{
 				Spec: pipelinev1.TaskRunSpec{
 					Params: nil,
@@ -158,7 +157,6 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 		})
 
 		It("should return error when TaskRun has no parameters", func() {
-			// Test case 8: TaskRun with empty parameters
 			tr := &pipelinev1.TaskRun{
 				Spec: pipelinev1.TaskRunSpec{
 					Params: []pipelinev1.Param{},
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -23,8 +23,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const systemNamespace = "multi-platform-controller"
-const userNamespace = "default"
+const (
+	systemNamespace = "multi-platform-controller"
+	userNamespace   = "default"
+)
 
 var cloudImpl MockCloud = MockCloud{Instances: map[cloud.InstanceIdentifier]MockInstance{}}
 
@@ -39,7 +41,6 @@ func setupClientAndReconciler(objs []runtimeclient.Object) (runtimeclient.Client
 }
 
 var _ = Describe("TaskRun Reconciler Tests", func() {
-
 	Describe("Test Config Map Parsing", func() {
 		var reconciler *ReconcileTaskRun
 
@@ -105,7 +106,7 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 	Describe("Test extractPlatform function", func() {
 		// Success cases first
 		It("should extract platform from TaskRun parameters successfully", func() {
-			// Test case 1: TaskRun with PLATFORM parameter
+			// Test case 1: TaskRun with PlatformParam parameter
 			tr := &pipelinev1.TaskRun{
 				Spec: pipelinev1.TaskRunSpec{
 					Params: []pipelinev1.Param{
@@ -118,7 +119,7 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 		})
 
 		It("should extract platform from TaskRun with multiple parameters", func() {
-			// Test case 2: TaskRun with multiple parameters including PLATFORM
+			// Test case 2: TaskRun with multiple parameters including PlatformParam
 			tr := &pipelinev1.TaskRun{
 				Spec: pipelinev1.TaskRunSpec{
 					Params: []pipelinev1.Param{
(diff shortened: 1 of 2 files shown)


pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
 		})
 	})
 
+	Describe("Test extractPlatform function", func() {
+		// Success cases first
+		It("should extract platform from TaskRun parameters successfully", func() {
+			// Test case 1: TaskRun with PLATFORM parameter
```
meyrevived, line 108: ExtraPlatform is looking for the platform parameter like so:
`if p.Name == PlatformParam {`
It does not look for _PLATFORM_

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+			// Test case 3: TaskRun with PLATFORM parameter but empty value
+			tr := &pipelinev1.TaskRun{
+				Spec: pipelinev1.TaskRunSpec{
+					Params: []pipelinev1.Param{
+						{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues("")},
+					},
+				},
+			}
+
+			Expect(extractPlatform(tr)).To(Equal(""))
```
meyrevived, line 145: This is a negative/sad path test case, it should have these assertions, similar to those of other sad paths test cases written below:
```
platform, err := extractPlatform(tr)
Expect(err).To(MatchError("failed to determine platform"))
Expect(platform).To(Equal(""))
```

I know it's not currently in the code, but it should be. It's just adding one line in extractPlatform, after it's been established that there is a parameter with the name PlatformParam.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+			testCases := []string{
+				"linux/amd64",
+				"linux/arm64",
+				"linux/s390x",
+				"linux/ppc64le",
+				"darwin/amd64",
+				"darwin/arm64",
+				"windows/amd64",
+			}
+
```
meyrevived, line 159: Please use a DescribeTable() here, which is the standard was of running the same assertions on multiple values. Also,  add [annotations](https://onsi.github.io/gomega/#annotating-assertions) to the assertion line to printf the failed platforms, if there are any.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+							{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues(expectedPlatform)},
+						},
+					},
+				}
+
+				Expect(extractPlatform(tr)).To(Equal(expectedPlatform))
+			}
+		})
+
+		It("should handle case sensitivity correctly", func() {
```
meyrevived, line 173: There is no code in extractPlatform that checks for case sensitivity, it's a feature of golang's == operator and nothing we should be testing. Also, the way the test is written will make it pass not because of any test case sensitivity, but simply because the other values are just "platform", not "Platform_Param_"

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+					Params: []pipelinev1.Param{
+						{Name: "OTHER_PARAM", Value: *pipelinev1.NewStructuredValues("other_value")},
+						{Name: "ANOTHER_PARAM", Value: *pipelinev1.NewStructuredValues("another_value")},
+					},
+				},
+			}
+
+			platform, err := extractPlatform(tr)
+			Expect(err).To(MatchError("failed to determine platform"))
+			Expect(platform).To(Equal(""))
```
sadlerap, line 219: All of the error cases can probably be simplified:

```suggestion
			Expect(extractPlatform(tr)).Error().To(MatchError("failed to determine platform"))
```

This asserts the error is as expected and all other return values are their zero-initialized value.  For more details, see [here in gomega's documentation][1].

[1]: https://onsi.github.io/gomega/#handling-errors

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+						Params: []pipelinev1.Param{
+							{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues(expectedPlatform)},
+						},
+					},
+				}
+
+				Expect(extractPlatform(tr)).To(Equal(expectedPlatform))
+			}
+		})
+
```
meyrevived, line 172: This is how ginkgo handles a case of multiple values that have the same assertion for the same testing purpose:

```suggestion
				DescribeTable("should extract platform with various common platform values",
			func(expectedPlatform string) {
				// Test case 4: Test with different common platform values
				tr := &pipelinev1.TaskRun{
					Spec: pipelinev1.TaskRunSpec{
						Params: []pipelinev1.Param{
							{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues(expectedPlatform)},
						},
					},
				}
				Expect(extractPlatform(tr)).To(Equal(expectedPlatform))
			},
			Entry("platform is linux/amd64", "linux/amd64"),
			Entry("platform is linux/arm64", "linux/arm64"),
			Entry("platform is linux/s390x", "linux/s390x"),
			Entry("platform is linux/ppc64le", "linux/ppc64le"),
			Entry("platform is darwin/amd64", "darwin/amd64"),
			Entry("platform is darwin/arm64", "darwin/arm64"),
			Entry("platform is windows/amd64", "windows/amd64"),
		)
```

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -102,6 +102,150 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+		It("should handle case sensitivity correctly", func() {
+			// Test case 5: Verify that parameter name matching is case sensitive
+			tr := &pipelinev1.TaskRun{
+				Spec: pipelinev1.TaskRunSpec{
+					Params: []pipelinev1.Param{
+						{Name: "platform", Value: *pipelinev1.NewStructuredValues("linux/amd64")},    // lowercase
+						{Name: "Platform", Value: *pipelinev1.NewStructuredValues("linux/arm64")},    // mixed case
+						{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues("linux/s390x")}, // uppercase (correct)
```
meyrevived, line 180: If we're focusing on case sensitivity, then what's being tested  is golang's `==` operator, not anything that has to do with the logic that the code of `extractPlatform`.
Since PlatformParam is a constant in taskrun.go and not something that's coming from outside of MPC (like Tetkon objects's field values, or something from AWS/IBMC's APIs) testing case sensitivity is a bit of a specific thing. 
We're basically monitoring any changes in the value of a const field in our own code. That's perfectly OK, it's something to sometimes is done in QE to raise a developer's awareness about changes they made. But if we're doing that,  why focus on case sensitivity? 
Testing if `const PlatformParam  = "PLATFORM"` is still a thing is basically done in Test case 7, so this test case is not useful.

meyrevived (changes requested):

meyrevived:
What should also be tested is validation for the allowed list of platforms. Currently, the PLATFORM parameter can contain anything, and there's nothing to stop `extractPLatform` from passing it on. 
We can fix it _temporarily_ with a slice of strings containing the current platforms and add an `&&` condition using golang's [slices.Contains](https://pkg.go.dev/slices#example-Contains) to make sure the value of the `PLATFORM` parameter is allowed. But we really have to start working on [KONFLUX-8416](https://issues.redhat.com/browse/KONFLUX-8416)

pkg/reconciler/taskrun/taskrun_test.go
filariow: Thanks for pointing this out. I'd prefer to punt this to a different task, so that in this one we just add missing test cases for current logic, in the other one we'll change the behavior of the function.

meyrevived: Sure. I plan on going over all the quick-fix bugs that I found during test writing, and fixing them. This is a good candidate, please open a jira bug about it and give it the `MPC-bugs` label so it'll be easily found.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -132,91 +130,43 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
 
-			Expect(extractPlatform(tr)).To(Equal("linux/amd64")) // Should return the first occurrence
+			Expect(extractPlatform(tr)).To(Equal(""))
 		})
 
-		// Error cases second
-		It("should return error when PLATFORM parameter is missing", func() {
-			// Test case 7: TaskRun without PLATFORM parameter
+		It("should return error when TaskRun has nil parameters", func() {
+			// Test case 9: TaskRun with nil parameters
```
filariow, line 162:
```suggestion
```

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -132,91 +130,43 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
-						{Name: "MIDDLE_PARAM", Value: *pipelinev1.NewStructuredValues("middle_value")},
-						{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues("linux/arm64")}, // Second occurrence
-						{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues("linux/s390x")}, // Third occurrence
+						{Name: PlatformParam, Value: *pipelinev1.NewStructuredValues("")},
 					},
 				},
 			}
 
-			Expect(extractPlatform(tr)).To(Equal("linux/amd64")) // Should return the first occurrence
+			Expect(extractPlatform(tr)).To(Equal(""))
```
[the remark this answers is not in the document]
filariow: makes sense, let's remove it

PR #492, opened 2025-07-09
Add conflict-prone TaskRun update function and use it whenever possible

mshaposhnik:
In previuos PR, I have deleted some intermediate TaskRun updates, because each update schedules the new reconcilliation, right, and we don't neeed this, since they seems may start in parallel (yes documentation says no but in practice we can still see it) 

but, TaskRun is not exclusive for MPC, so tekton controller also updating them

so, now when the code reaches to the final very important TR update - when the instance ID and/or IP set, it often catches the conflict exception. and this i info is missing. 

so all starts over again, but the VM provisioned in 1st turn become an zombie..

This PR introduces the conflict-resillient update fuction

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -979,7 +979,7 @@ func UpdateTaskRunWithRetry(ctx context.Context, client client.Client, apiReader
 		return updateTaskRun(ctx, client, apiReader, tr)
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update TaskRun after %d attempts, last error: %w", retry.DefaultRetry.Steps, err)
+		return fmt.Errorf("failed to update TaskRun, last error: %w", err)
 	}
 	return nil
 }
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/client-go/util/retry"
+
 	"knative.dev/pkg/kmeta"
 
 	mpcmetrics "github.com/konflux-ci/multi-platform-controller/pkg/metrics"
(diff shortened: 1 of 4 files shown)


filariow (changes requested):
reviewing tests

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -970,3 +970,104 @@ type Host struct {
 func platformLabel(platform string) string {
 	return strings.ReplaceAll(platform, "/", "-")
 }
+
+// UpdateTaskRunWithRetry performs a conflict-resilient update of a TaskRun object.
+// On conflict, it fetches the latest version and merges labels, annotations, and finalizers from the incoming TaskRun.
+func UpdateTaskRunWithRetry(ctx context.Context, client client.Client, apiReader client.Reader, tr *tektonapi.TaskRun, maxRetries int) error {
```
filariow, line 976: We could use the [retry logic](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict) available from the client-go library. It seems to me that we could simplify our code and offload the implementation of the retry logic to the kubernetes toolchain. Do you think it is a good fit here?

filariow: assuming we move the update logic to a function called `updateTaskRun` it become something like follows:

```golang
func UpdateTaskRunWithRetry(ctx context.Context, client client.Client, apiReader client.Reader, tr *tektonapi.TaskRun, maxRetries int) error {
	  return retry.RetryOnConflict(retry.DefaultRetry, func() error { 
	  	  return updateTaskRun(ctx, client, apiReader, tr)
	  }
}
```

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -707,6 +716,216 @@ var _ = Describe("TaskRun Reconciler Tests", func() {
...
+
+			// Verify the update was applied
+			updated := &pipelinev1.TaskRun{}
+			Expect(client.Get(ctx, types.NamespacedName{Namespace: tr.Namespace, Name: tr.Name}, updated)).To(Succeed())
+			Expect(updated.Labels).To(HaveKeyWithValue("conflict-label", "conflict-value"))
+			Expect(updated.Annotations).To(HaveKeyWithValue("conflict-annotation", "conflict-value"))
+			Expect(updated.Finalizers).To(ContainElement("conflict-finalizer"))
+		})
+
+		It("should merge labels and annotations correctly after conflict", func(ctx SpecContext) {
```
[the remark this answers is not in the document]
mshaposhnik: i will commit the test if you know how to solve it please do

mshaposhnik:
#494  was merged istead

PR #497, opened 2025-07-16
proposal for acceptance tests with chainsaw

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>


gbenhaim:
What would be the benefit of using Chainsaw over writing the tests in Golang and use Ginkgo (where we have linting, auto completion, AI that can assist us) ?

PR #499, opened 2025-07-28
Make MPC running_tasks and waiting_tasks metrics to contain source namespace

glevi-rh:
## What
Converted the `RunningTasks` and `WaitingTasks` metrics from `Gauge` to **`GaugeVec`**. This change includes all necessary code updates to support the new metric type.

## Why
To enable the addition of a **`source_namespace` label** to `multi_platform_controller_running_tasks` and `multi_platform_controller_waiting_tasks`. This label will significantly improve Konflux support's ability to monitor resource utilization, allowing them to easily identify which tenants are consuming the most VM resources or experiencing the longest wait times.

### DIFF
--- a/pkg/metrics/platform_test.go
+++ b/pkg/metrics/platform_test.go
@@ -21,28 +21,28 @@ var _ = Describe("PlatformMetrics", func() {
 			Expect(RegisterPlatformMetrics(ctx, platform)).NotTo(HaveOccurred())
 			//resetting counters
 			HandleMetrics(platform, func(m *PlatformMetrics) {
-				m.RunningTasks.Set(0)
+				m.RunningTasks.WithLabelValues(platform, "test-namespace").Set(0)
 			})
 			HandleMetrics(platform, func(m *PlatformMetrics) {
-				m.WaitingTasks.Set(0)
+				m.WaitingTasks.WithLabelValues(platform, "test-namespace").Set(0)
 			})
 
 		})
 		When("When appropriate condition happened", func() {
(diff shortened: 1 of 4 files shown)


enkeefe00 (approved):

pkg/metrics/platform_test.go
```diff
@@ -150,16 +150,28 @@ var _ = Describe("PlatformMetrics", func() {
...
+						if l.GetName() == "platform" && l.GetValue() == platform {
+							hasPlatform = true
+						}
+						if l.GetName() == "namespace" && l.GetValue() == namespace {
+							hasNamespace = true
+						}
+					}
+					if hasPlatform && hasNamespace {
+						return int(m.Gauge.GetValue()), nil
+					}
```
filariow, line 174: minor: idea to simplify this block:

```suggestion
				if m.Gauge == nil {
					continue
				}

				if hasLabel(m.Label, "platform", platform) && hasLabel(m.Label, "namespace", namespace) {
					return int(m.Gauge.GetValue()), nil
				}
```

where `hasLabel` is as follow:

```golang
import io_prometheus_client "github.com/prometheus/client_model/go"

func hasLabel(lp []*io_prometheus_client.LabelPair, name, value string) bool {
	return slices.ContainsFunc(lp, func(l *io_prometheus_client.LabelPair) bool {
		return l.GetName() == name && l.GetValue() == value
	})
}
```

filariow (commented):
lgtm, just a minor feedback

meyrevived (approved):
/lgtm

PR #502, opened 2025-08-03
Handle platform label mismatch

glevi-rh:
## what
Added `platformLabelComperator` and `platformLabel` functions
## why
In some tests ("Provision Failure" test among them) the platform variable contains the platform name with '-' (dash sign) and the key in `platformMetrics` is the platform name with a '/' (slash). The new functions handling that kind of mismatch.

### DIFF
--- a/pkg/metrics/platform.go
+++ b/pkg/metrics/platform.go
@@ -2,6 +2,7 @@ package mpcmetrics
 
 import (
 	"context"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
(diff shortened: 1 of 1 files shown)


pkg/metrics/platform.go
```diff
@@ -117,8 +118,24 @@ func RegisterPlatformMetrics(_ context.Context, platform string) error {
...
+}
+
+// Compare the platform label to the format used by PlatformMetrics and return the corresponding PlatformMetrics
+func platformLabelComperator(platformMetrics map[string]*PlatformMetrics, platform string) *PlatformMetrics {
+	if pmetrics, ok := platformMetrics[platform]; ok {
+		return pmetrics
+	}
+	if pmetrics, ok := platformMetrics[platformLabel(platform)]; ok {
+		return pmetrics
+	}
```
meyrevived, line 133: I think this is a smoother way of doing it, while keeping the safety of fetching values from a map the way you did. WDYT?

```suggestion
	pmetrics, ok := platformMetrics[platform]

	if !ok {
		pmetrics, ok = platformMetrics[platformLabel(platform)]
	}
	if ok {
		return pmetrics
	}
```

meyrevived (commented):
I like the safe middle ground between the way platform names are actually written and Stuart's function.
Adding just a small style and readability improvement suggestion.

meyrevived (approved):
/lgtm

PR #506, opened 2025-08-11
Waiting tasks bug fix

glevi-rh:
### What

* **Removed the line that deletes the `WaitingForPlatformLabel` label within the `handleWaitingTasks` function.** This was previously causing the label to be deleted prematurely.
* **Updated the unit tests to check for changes to the `WaitingTasks` counter.** The tests now ensure that the counter is both incremented and decremented correctly within the `handleHostAllocation` function.

### Why

* **To ensure the `WaitingForPlatformLabel` label is only removed after a host has been successfully allocated.** This guarantees that a task remains in the "waiting" state until a valid host is assigned.
* **To fix a bug where the `WaitingTasks` counter was being incorrectly managed.** Previously, the counter was only incremented, leading to a misrepresentation of the number of waiting tasks. This change ensures the counter accurately reflects the current state by correctly incrementing and decrementing the count as tasks enter and leave the waiting state.


meyrevived (approved):
/lgtm

mshaposhnik (approved):

PR #509, opened 2025-08-21
KFLUXINFRA-1895 work - validate the value of TaskRun's PLATFORM parameter

meyrevived:
First commit:
1. Added back testing specs that Francesco wrote for `extractPlatform` and got wiped out by the Big Bad Refactoring.
2. Added validation to the PLATFORM value string according the  acceptance criteria - enforces "label/label" structure with RFC 1035,  compliance and only accept allowed platforms (list of platforms taken from host-config.yaml files of all of our clusters).
3. Updates the tests for extractPlatform to include tests for the new validation capabilities.

Assisted-by: Claude Code


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -89,6 +90,48 @@ const (
...
+// RFC 1035 label name regex: starts with letter, contains only lowercase letters/numbers/hyphens, max 63 chars, ends with alphanumeric
+var rfc1035LabelRegex = regexp.MustCompile(`^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`)
+
+var allowedPlatformsMap = func() map[string]struct{} {
+	m := make(map[string]struct{}, len(allowedPlatforms))
+	for _, platform := range allowedPlatforms {
+		m[platform] = struct{}{}
+	}
+	return m
+}()
```
filariow, line 134: There is something similar in "k8s.io/utils/set" that we can reuse

```suggestion
import "k8s.io/utils/set"

var allowedPlatformsMap = func() set.Set {
  return set.New(allowedPlatforms...)
}()
```

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -89,6 +90,48 @@ const (
...
+	"linux-d320-c4xlarge/amd64", "linux-d320-c4xlarge/arm64",
+	// s390x platforms
+	"linux/s390x", "linux-large/s390x", "linux-d200/s390x", "linux-d200-large/s390x",
+	// ppc64le platforms
+	"linux/ppc64le", "linux-large/ppc64le", "linux-xlarge/ppc64le", "linux-2xlarge/ppc64le", "linux-d200-large/ppc64le",
+	"linux-d200-xlarge/ppc64le", "linux-d200-2xlarge/ppc64le",
+}
+
+// RFC 1035 label name regex: starts with letter, contains only lowercase letters/numbers/hyphens, max 63 chars, ends with alphanumeric
+var rfc1035LabelRegex = regexp.MustCompile(`^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`)
```
filariow, line 126: I'd suggest relying on [k8s.io/apimachinery/pkg/util/validation's IsDNS1035Label](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1035Label). It will reduce the logic we have to maintain.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -469,7 +512,33 @@ func (r *ReconcileTaskRun) handleUserTask(ctx context.Context, tr *tektonapi.Tas
...
+			parts := strings.Split(platform, "/")
+			if len(parts) != 2 {
+				return "", errInvalidPlatformFormat
+			}
+			// Validate each part against RFC 1035 label name rules
+			for _, part := range parts {
+				if !rfc1035LabelRegex.MatchString(part) {
+					return "", errInvalidPlatformFormat
+				}
+			}
```
sadlerap, line 536: Why do we need to do custom parsing here?  Why isn't checking membership in the allowed platforms map sufficient?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -89,6 +90,48 @@ const (
...
+	"linux-d160-c4xlarge/amd64", "linux-d160-c4xlarge/arm64", "linux-d160-c8xlarge/amd64", "linux-d160-c8xlarge/arm64",
+	"linux-d160-m8-8xlarge/arm64", "linux-d160-m7-8xlarge/amd64",
+	// D320-series instances
+	"linux-d320-c4xlarge/amd64", "linux-d320-c4xlarge/arm64",
+	// s390x platforms
+	"linux/s390x", "linux-large/s390x", "linux-d200/s390x", "linux-d200-large/s390x",
+	// ppc64le platforms
+	"linux/ppc64le", "linux-large/ppc64le", "linux-xlarge/ppc64le", "linux-2xlarge/ppc64le", "linux-d200-large/ppc64le",
+	"linux-d200-xlarge/ppc64le", "linux-d200-2xlarge/ppc64le",
+}
```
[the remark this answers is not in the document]
sadlerap: agreed, this should likely be pulled from the host config rather than hardcoded.

pkg/reconciler/taskrun/taskrun.go
meyrevived: Also, Francesco specified acceptance criteria in the task.

sadlerap: I still don't think this is right.

1. This is an optimization for the case where the build platform we get is bad.  Do we run into it much in practice?
2. Go's maps (which is how the data structures in `k8s.io/utils/set` work) should have key lookups be pretty fast.  Do we have performance metrics showing this is a bottleneck?  If so, how much faster is it to do a regex evaluation beforehand?

@filariow could you clarify why this was a part of the acceptance criteria?  I'm not quite understanding the need here.

meyrevived: I literally had a [whole talk about the importance of data validation](https://docs.google.com/presentation/d/1CGp3eprxpr-hg_hpLQDXq82fFStZypMHsfONVke9nMQ/edit?usp=drivesdk). I put a lot of effort into that crappy AI generated artwork  😭😭

meyrevived:
Replaced with a bigger effort

PR #510, opened 2025-08-26
Running tasks counter bug fix

glevi-rh:
### What

* Modified the `handleHostAssigned()` function to improve how it updates the TaskRun with cleanup changes.

* Moved the decrement of the `multi_platform_controller_running_tasks` metric to occur only after the TaskRun has been successfully updated.

### Why

* Addresses a race condition that frequently occurs between the MPC controller and the Tekton controller when a TaskRun completes. The new update function now merges annotations and labels to prevent conflicts and ensure a successful TaskRun update.

* Prevents potential multiple decrements of the `multi_platform_controller_running_tasks` counter. Previously, the counter was decremented before a successful update, which could lead to multiple decrements if the cleanup process required retries.

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -595,14 +595,6 @@ func (r *ReconcileTaskRun) handleHostAssigned(ctx context.Context, tr *tektonapi
 		return reconcile.Result{}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


enkeefe00 (approved):
I approve with Francesco's changes implemented.

filariow (approved):
/lgtm

PR #527, opened 2025-09-06
make: invoke ginkgo from `go run`

sadlerap:
Invoke ginkgo through `go run`, rather than directly as an external binary.  This has two advantages:
- Reduces our external binary dependency count by one, making onboarding of new developers easier.
- Invoked ginkgo binaries will no longer become out-of-sync with the version MPC depends on.  This is useful for developers working on multiple projects, each of which may have their own version of ginkgo.


meyrevived:
1. You'll need to first merge #458 into this to allow it to merge
2. Do you want to back this up for the makefile by adding tools.go?

PR #539, opened 2025-09-10
KFLUXINFRA-2233 work: Validate the format of the user-provided PLATFORM string value in a TaskRun

meyrevived:
This PR includes the fix for the problem that started this whole mess - nothing verifies that a platform listen in a user TaskRun's PLATFORM parameter is actually valid. 
Validation for the [host platform field](https://github.com/konflux-ci/multi-platform-controller/pull/491#discussion_r2185914732):  label/label, where a label is respecting the [RFC 1035 format.](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names). This is done in a format validation method.
extractPlatform has been moved to the new validation code file, and is not longer directly called by taskrun.go's handleUserTask. Instead, a new method which both checks the PLATFORM parameter even exists and can be extracted, and then validates the format of PLATFORM's string value.

### DIFF
--- a/pkg/reconciler/taskrun/validation.go
+++ b/pkg/reconciler/taskrun/validation.go
@@ -1,0 +1,87 @@
+package taskrun
+
+import (
+	"strings"
+
+	errors2 "github.com/pkg/errors"
+	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+var (
+	errInvalidPlatformFormat    = errors2.New("platform must be in format 'label/label' where each label follows Kubernetes RFC 1035 label name format")
+	errMissingPlatformParameter = errors2.New("PLATFORM parameter not found in TaskRun parameters")
+)
+
+// validatePlatformFormat validates a platform string according to the controller's rules
+// Validation rules:
+// - Exceptional platforms: "local", "localhost", "linux/x86_64" bypass format validation
+// - Standard platforms: must be "label/label" format where each label follows RFC 1035
+// - RFC 1035 compliance: lowercase alphanumeric characters and hyphens, start/end with alphanumeric
+//
+// Returns:
+// - nil if platform is valid
+// - errInvalidPlatformFormat if format is incorrect
+func validatePlatformFormat(platform string) error {
+	// Check for exceptional platforms that bypass standard validation
+	if platform == "local" || platform == "localhost" || platform == "linux/x86_64" {
+		return nil
+	}
+
+	// Validate platform format: must be "label/label" where each label follows RFC 1035
+	parts := strings.Split(platform, "/")
+	if len(parts) != 2 {
+		return errInvalidPlatformFormat
+	}
+
+	// Validate each part against RFC 1035 label name rules
+	for _, part := range parts {
+		if p := validation.IsDNS1035Label(part); len(p) != 0 {
+			return errInvalidPlatformFormat
+		}
+	}
+
+	return nil
+}
+
+// validatePlatform extracts and validates the platform parameter from a TaskRun
+// This function combines platform extraction and format validation in a single operation.
+// It first extracts the PLATFORM parameter from the TaskRun, then validates its format.
+//
+// Parameters:
+// - tr: The TaskRun object to extract and validate the platform from
+//
+// Returns:
+// - string: The validated platform value if found and valid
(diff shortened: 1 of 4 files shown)


pkg/reconciler/taskrun/validation_test.go
```diff
@@ -0,0 +1,143 @@
...
+	// This section tests the combined validation function
+	Describe("The validatePlatform function", func() {
+
+		When("the TaskRun contains a valid platform parameter", func() {
+			DescribeTable("it should return the platform and no error",
+				func(platformValue string) {
+					tr := createTrWithPlatform(&platformValue)
+					platform, err := validatePlatform(tr)
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(platform).Should(Equal(platformValue))
```
sadlerap, line 124:
```suggestion
					Expect(validatePlatform(tr)).To(Equal(platformValue))
```

pkg/reconciler/taskrun/validation_test.go
```diff
@@ -0,0 +1,143 @@
...
+				Entry("for the 'linux/x86_64' exception", "linux/x86_64"),
+			)
+		})
+
+		When("the TaskRun contains an invalid or missing platform parameter", func() {
+			DescribeTable("it should return the corresponding error",
+				func(platformValue *string, expectedError error) {
+					tr := createTrWithPlatform(platformValue)
+					_, err := validatePlatform(tr)
+					Expect(err).Should(MatchError(expectedError))
```
sadlerap, line 136:
```suggestion
					Expect(validatePlatform(tr)).Error().To(MatchError(expectedError))
```

This can be applied to a few other places in this PR.

pkg/reconciler/taskrun/validation.go
```diff
@@ -0,0 +1,87 @@
+package taskrun
+
+import (
+	"strings"
+
+	errors2 "github.com/pkg/errors"
```
sadlerap, line 6: two questions:
- Why rename this package during imports?  It doesn't conflict with any other import in this file.
- This package has been archived since 2021 and appears to be deprecated.  Should we consider removing its use from MPC?  Removal doesn't necessarily need to happen in this PR, but we shouldn't introduce new usage of it if we want to go that path.

pkg/reconciler/taskrun/validation_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	pipelinev1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+)
+
+// A helper to create a TaskRun with a platform parameter.
+// It accepts a pointer to a string so we can pass 'nil' to simulate a missing parameter.
+func createTrWithPlatform(platform *string) *pipelinev1.TaskRun {
```
sadlerap, line 14: We only use this behavior in one specific test, and I'm not a fan of pointers to types to make behavior conditional like this.

Consider special-casing that test out from its test fixture, and don't call this function in its body to make the platform parameter conditional.  We might also change the interface of this function to take a taskrun and return its modified version (or take a pointer and modify it without a copy).

pkg/reconciler/taskrun/validation_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+		When("the TaskRun contains an invalid or missing platform parameter", func() {
+			DescribeTable("it should return the corresponding error",
+				func(platformValue *string, expectedError error) {
+					tr := createTrWithPlatform(platformValue)
+					_, err := validatePlatform(tr)
+					Expect(err).Should(MatchError(expectedError))
+				},
+				Entry("when the parameter is missing", nil, errMissingPlatformParameter),
+				Entry("when the parameter format is invalid", &[]string{"koko_hazamar/moshe_ata_lo_kipod"}[0], errInvalidPlatformFormat),
```
sadlerap, line 139: We should consider adding a test for taskruns that have multiple platform parameters set.  They should be rejected as invalid, since the build platform is ambiguous.

pkg/reconciler/taskrun/validation_test.go
meyrevived: tl;dr: sheer and careless stupidity.
Re-added to the next commit.

pkg/reconciler/taskrun/validation_test.go
meyrevived: tl;dr: we used to have them, they fell victim to ~~The Great Refactoring~~ sheer and careless stupidity.
Re-added in the latest commit.

PR #546, opened 2025-09-16
ci: use a better golangci-lint setup

sadlerap:
This changes two things about how golangci-lint is run in CI:
- `golangci-lint` emits sarif reports, which are uploaded to github.  This enables lint warnings to appear in PRs as comments, rather than requiring us to look at checkrun logs.
- Pin the version of golangci-lint to what's in the modfile.  This prevents the version running in CI from desynchronizing with what we run locally, and it allows what we use in CI to get updated by renovate.


filariow (approved):
/lgtm

PR #569, opened 2025-10-05
KFLUXINFRA-2234 first commit - moving to reading host configuration data and validating it before returning it to readConfigurations

meyrevived:
Committing the initial work  for the task:
1. Introduces typed structs for the various types of hosts available in host-config.yaml along with a full validation function for the configurations in it.
2. Refactor taskrun->readConfiguration to use this new, safe model instead of a raw map.
3. Tests for all new functions and functionalities.

Assisted-by: Claude Code

### DIFF
--- a/pkg/reconciler/taskrun/config.go
+++ b/pkg/reconciler/taskrun/config.go
@@ -1,0 +1,391 @@
+package taskrun
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	// Default allocation timeout for dynamic platforms in seconds (10 minutes)
+	defaultAllocationTimeout = 600
+)
+
+// parsePlatformList parses a comma-separated list of platforms, validates each one, and returns a slice of valid
+// platforms. Handles trailing commas.
+func parsePlatformList(platformList string, platformType string) ([]string, error) {
+	if platformList == "" {
+		return []string{}, nil
+	}
+
+	platforms := strings.Split(platformList, ",")
+	result := []string{}
+	for i, platform := range platforms {
+		platform = strings.TrimSpace(platform)
+		if platform == "" {
+			// Allow empty strings only at the end (from trailing commas)
+			if i == len(platforms)-1 {
+				continue
+			}
+			return nil, fmt.Errorf("invalid %s platform '': empty platform in middle of list", platformType)
+		}
+		if err := validatePlatformFormat(platform); err != nil {
+			return nil, fmt.Errorf("invalid %s platform '%s': %w", platformType, platform, err)
+		}
+		result = append(result, platform)
+	}
+	return result, nil
+}
+
+// ClusterHostConfig represents the complete host configuration from ConfigMap
+type ClusterHostConfig struct {
+	LocalPlatforms         []string
+	DynamicPlatforms       map[string]DynamicPlatformConfig
+	DynamicPoolPlatforms   map[string]DynamicPoolPlatformConfig
+	StaticPlatforms        map[string]StaticHostConfig
+	DefaultInstanceTag     string
+	AdditionalInstanceTags map[string]string
+}
+
+// DynamicPlatformConfig holds configuration for a single dynamic platform
+type DynamicPlatformConfig struct {
+	Type              string `mapstructure:"type"`
+	MaxInstances      int    `mapstructure:"max-instances"`
+	InstanceTag       string `mapstructure:"instance-tag,omitempty"`
+	AllocationTimeout int64  `mapstructure:"allocation-timeout,omitempty"`
+	SSHSecret         string `mapstructure:"ssh-secret"`
+	SudoCommands      string `mapstructure:"sudo-commands,omitempty"`
+}
+
+// DynamicPoolPlatformConfig holds configuration for a single dynamic platform in a host pool
+type DynamicPoolPlatformConfig struct {
+	Type         string `mapstructure:"type"`
+	MaxInstances int    `mapstructure:"max-instances"`
+	Concurrency  int    `mapstructure:"concurrency"`
+	MaxAge       int    `mapstructure:"max-age"` // in minutes
+	InstanceTag  string `mapstructure:"instance-tag,omitempty"`
+	SSHSecret    string `mapstructure:"ssh-secret"`
+}
+
+// StaticHostConfig represents a single static host configuration
+type StaticHostConfig struct {
+	Address     string `mapstructure:"address"`
+	User        string `mapstructure:"user"`
+	Platform    string `mapstructure:"platform"`
+	Secret      string `mapstructure:"secret"`
+	Concurrency int    `mapstructure:"concurrency"`
+}
+
+// ParseAndValidate parses and validates host configuration data from a ConfigMap
+// This function extracts configuration for all platform types (local, dynamic, dynamic-pool, static)
+// and validates each field according to the controller's validation rules.
+//
+// Parameters:
+// - ctx: Context for AWS resource validation
+// - kubeClient: Kubernetes client for AWS resource validation
+// - data: The ConfigMap data map containing host configuration
+// - systemNamespace: System namespace for AWS resource validation
+//
+// Returns:
+// - *ClusterHostConfig: The parsed and validated configuration
+// - error: Validation error if any field fails validation
+func ParseAndValidate(ctx context.Context, kubeClient client.Client, data map[string]string, systemNamespace string) (*ClusterHostConfig, error) {
+	config := &ClusterHostConfig{
+		LocalPlatforms:         []string{},
+		DynamicPlatforms:       make(map[string]DynamicPlatformConfig),
+		DynamicPoolPlatforms:   make(map[string]DynamicPoolPlatformConfig),
+		StaticPlatforms:        make(map[string]StaticHostConfig),
+		AdditionalInstanceTags: make(map[string]string),
+	}
+
+	// Parse local platforms
+	localPlatforms, err := parsePlatformList(data[LocalPlatforms], "local")
+	if err != nil {
+		return nil, err
+	}
+	config.LocalPlatforms = localPlatforms
+
+	// Parse dynamic platforms
+	dynamicPlatforms, err := parsePlatformList(data[DynamicPlatforms], "dynamic")
+	if err != nil {
+		return nil, err
+	}
+	for _, platform := range dynamicPlatforms {
+		platformConfigName := strings.ReplaceAll(platform, "/", "-")
+		prefix := "dynamic." + platformConfigName + "."
+
+		// Parse and validate dynamic platform configuration
+		dynamicConfig := DynamicPlatformConfig{}
+
+		// Type field (required)
+		if typeName := data[prefix+"type"]; typeName != "" {
+			dynamicConfig.Type = typeName
+		} else {
+			return nil, fmt.Errorf("dynamic platform '%s': type field is required", platform)
+		}
+
+		// Max instances (required)
+		if maxInstancesStr := data[prefix+"max-instances"]; maxInstancesStr != "" {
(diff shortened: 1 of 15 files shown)


pkg/reconciler/taskrun/config.go
```diff
@@ -0,0 +1,391 @@
...
+	DynamicPlatforms       map[string]DynamicPlatformConfig
+	DynamicPoolPlatforms   map[string]DynamicPoolPlatformConfig
+	StaticPlatforms        map[string]StaticHostConfig
+	DefaultInstanceTag     string
+	AdditionalInstanceTags map[string]string
+}
+
+// DynamicPlatformConfig holds configuration for a single dynamic platform
+type DynamicPlatformConfig struct {
+	Type              string `mapstructure:"type"`
```
filariow, line 54: do we need the `mapstructure` annotation? I don't see mapstructure used for parsing the configuration.

pkg/reconciler/taskrun/config.go
```diff
@@ -0,0 +1,391 @@
...
+	if additionalTags := data[AdditionalInstanceTags]; additionalTags != "" {
+		tags := strings.Split(additionalTags, ",")
+		for _, tag := range tags {
+			tag = strings.TrimSpace(tag)
+			parts := strings.Split(tag, "=")
+			if len(parts) >= 2 {
+				config.AdditionalInstanceTags[parts[0]] = parts[1]
+			} else {
+				return nil, fmt.Errorf("invalid additional instance tag format '%s': must be key=value", tag)
+			}
```
filariow, line 386: if more than one `=` is found, the format is still invalid. We should return an error.

```suggestion
			if len(parts) != 2 {
				return nil, fmt.Errorf("invalid additional instance tag format '%s': must be key=value", tag)
			}
			config.AdditionalInstanceTags[parts[0]] = parts[1]
```

pkg/reconciler/taskrun/config.go
```diff
@@ -0,0 +1,391 @@
...
+// Parameters:
+// - ctx: Context for AWS resource validation
+// - kubeClient: Kubernetes client for AWS resource validation
+// - data: The ConfigMap data map containing host configuration
+// - systemNamespace: System namespace for AWS resource validation
+//
+// Returns:
+// - *ClusterHostConfig: The parsed and validated configuration
+// - error: Validation error if any field fails validation
+func ParseAndValidate(ctx context.Context, kubeClient client.Client, data map[string]string, systemNamespace string) (*ClusterHostConfig, error) {
```
filariow, line 94: this function is quite big, can we split it in multiple ones?

pkg/reconciler/taskrun/validation.go
```diff
@@ -85,3 +105,217 @@ func extractPlatform(tr *tektonapi.TaskRun) (string, error) {
...
+// Returns:
+// - int: The validated integer value
+// - error: errInvalidNumericValue if value is invalid or out of range
+func validateNumericValue(value string, maxValue int) (int, error) {
+	num, err := strconv.Atoi(value)
+	if err != nil || num != max(0, min(num, maxValue)) {
+		return 0, errInvalidNumericValue
+	}
+	return num, nil
+}
```
filariow, line 127: Some ideas on how we could improve this function:
* I'd suggest to not suppress the error returned by `strconv.Atoi`. We can join it via `errors.Join`.
* The `max(..., min..)` logic is quite obscure, we could simplify it
* We could improve the name of the function to make it clear it will parse and check that the number is `[1, maxValue]`

```suggestion
func parseNonZeroPositiveNumber(value string, maxValue int) (int, error) {
	num, err := strconv.Atoi(value)
	if err != nil {
		return -1, errors.Join(errInvalidNumericValue, err)
	}
	if num < 0 || num > maxValue {
		return -1, errInvalidNumericValue
	}
	return num, nil
}
```

pkg/reconciler/taskrun/validation.go
```diff
@@ -2,15 +2,35 @@ package taskrun
...
+	errIBMHostSecretEmpty            = errors.New("host secret value cannot be empty")
+	errIBMHostSecretPlatformMismatch = errors.New("host secret key and value must contain matching platform substring")
+
+	maxInstancesValue = 250
+	// Maximum allocation timeout value in seconds (20 minutes)
+	maxAllocationTimeout = 1200
+	// Maximum static host concurrency
+	maxStaticConcurrency = 8
+	// Maximum pool host age in minutes (24 hours)
+	maxPoolHostAge = 1440
```
filariow, line 33: how can we avoid hardcoding these values? If in future we want to change them, we'll need a new release. Also we'll not be able to have different values for different clusters.

pkg/reconciler/taskrun/validation.go
```diff
@@ -85,3 +105,217 @@ func extractPlatform(tr *tektonapi.TaskRun) (string, error) {
...
+// - errInvalidIPFormat if value is not a valid IP address
+// - dynamic error if IP is valid but host is unreachable (includes obfuscated IP)
+func validateIP(value string) error {
+	ip := net.ParseIP(value)
+	if ip != nil {
+		// IP format is valid, now check reachability
+		if err := ec2Helpers.PingIPAddress(value); err != nil {
+			obfuscatedIP := obfuscateIP(value)
+			return fmt.Errorf("IP address is valid but host %s is unreachable", obfuscatedIP)
+		}
```
filariow, line 180: If we perform this here, MPC will not be able to start if at least one machine is unreachable. I think it's valuable to validate the format, but I'd suggest not to check VM's reachability while parsing the config.

pkg/reconciler/taskrun/config.go
```diff
@@ -0,0 +1,391 @@
...
+					return nil, fmt.Errorf("dynamic platform '%s': ssh-secret cannot be empty for AWS platforms", platform)
+				}
+				dynamicConfig.SSHSecret = sshSecret
+			} else {
+				// IBM platforms: validate using validateIBMHostSecret
+				if err := validateIBMHostSecret(platform, sshSecret); err != nil {
+					return nil, fmt.Errorf("dynamic platform '%s': invalid ssh-secret '%s': %w", platform, sshSecret, err)
+				}
+				dynamicConfig.SSHSecret = sshSecret
+			}
```
filariow, line 175: We should not assume that if `dynamicConfig.Type` is not `aws` then it's IBM. Let's make sure the value meets our expectations.

```suggestion
			switch dynamicConfig.Type {
			case "aws":
				// AWS platforms: ensure the field has a non-empty value
				if strings.TrimSpace(sshSecret) == "" {
					return nil, fmt.Errorf("dynamic platform '%s': ssh-secret cannot be empty for AWS platforms", platform)
				}
				dynamicConfig.SSHSecret = sshSecret
			case "ibm":
				// IBM platforms: validate using validateIBMHostSecret
				if err := validateIBMHostSecret(platform, sshSecret); err != nil {
					return nil, fmt.Errorf("dynamic platform '%s': invalid ssh-secret '%s': %w", platform, sshSecret, err)
				}
				dynamicConfig.SSHSecret = sshSecret
			default:
			       return nil, fmt.Errorf("invalid type: expect 'ibm' or 'aws', got '%s', dynamicConfig.Type)
			}
```

pkg/reconciler/taskrun/validation.go
```diff
@@ -85,3 +105,217 @@ func extractPlatform(tr *tektonapi.TaskRun) (string, error) {
...
+	}
+	// Instance types must be an exact match.
+	if keyInstanceType != valueInstanceType {
+		return fmt.Errorf("instance type mismatch: key has '%s', value has '%s'", keyInstanceType, valueInstanceType)
+	}
+	return nil
+}
+
+// parseDynamicHostInstanceTypeKey extracts the platform and instance type from the platform config name.
+// This function parses the simplified platform config name format and normalizes multi-part instance types.
```
filariow, line 245: what do we mean by normalization here? which format is a normalized value respecting?

pkg/reconciler/taskrun/validation.go
```diff
@@ -85,3 +105,217 @@ func extractPlatform(tr *tektonapi.TaskRun) (string, error) {
...
+	// Platform is everything after the last dash.
+	platform = platformConfigName[lastDash+1:]
+	// Instance type is everything between the first and last dash.
+	instanceType = platformConfigName[firstDash+1 : lastDash]
+
+	// Normalize the instance type if it contains multiple parts
+	instanceParts := strings.Split(instanceType, "-")
+	if len(instanceParts) > 1 {
+		sort.Strings(instanceParts)
+		instanceType = strings.Join(instanceParts, "-")
```
filariow, line 278: why do we need to sort here?

filariow (changes requested):
Still going through this. May we split this PR in multiple ones?

meyrevived:
@filariow this task was already split in two to keep the size of the PR smaller. Half of the host configuration reading is done by the cloud providers' packages, and more validation code will be needed there. Some of the comments you left so far I can answer now, some someggestions I can commit already, and some will require me to actually sit in front of my laptop. Take your time, Allison well.

pkg/reconciler/taskrun/config.go
meyrevived: It's much smaller than taskrun.go -> readComfiguration where all of the parsing logic comes from.

pkg/reconciler/taskrun/config.go
meyrevived: Added to changes commit

pkg/reconciler/taskrun/validation.go
meyrevived: For some reason, the order of instance type and instance platform in the instance-tag key is not the same as they appear in the value - in the key it's `dynamic.linux-<instance type>-<platform>.instance-tag` but in the value for this key it's `<environment, or an additional part besides the environment>-<platform>-<instance type>`.  When the instance type is multi-parted itself (like `extra-fast` or `d160-mlarge`)  there needs to be some additional re-organization, as you can see documented in line 273.

pkg/reconciler/taskrun/validation.go
meyrevived: See the previous comment.

pkg/reconciler/taskrun/config_test.go
```diff
@@ -0,0 +1,747 @@
...
+			})
+
+			It("should return error when one tag in list is invalid", func(ctx SpecContext) {
+				data := map[string]string{
+					AdditionalInstanceTags: "env=production,invalid-tag,team=backend",
+				}
+
+				_, err := ParseAndValidate(ctx, nil, data, systemNamespace)
+				Expect(err.Error()).Should(ContainSubstring("invalid additional instance tag format 'invalid-tag'"))
+			})
```
enkeefe00, line 691: Can you add a test case fora  tag list like the following: `"env=production,invalid-tag-pair==value"`?

testing/sources/host-config.yaml
```diff
@@ -0,0 +1,654 @@
...
+  dynamic.linux-g6xlarge-amd64.user-data: |-
+    Content-Type: multipart/mixed; boundary="//"
+    MIME-Version: 1.0
+    
+    --//
+    I have a friend, he fried his brain backpacking through India
+    Thinks he's a hedgehog
+    Moshe - you are not a hedgehog
+    Moshe - you are not a hedgehog
+    Not a hedgehog, not a hedgehog, not a hedgehog
```
enkeefe00, line 653: Not a hedgehog :hedgehog:

enkeefe00 (commented):
Not in the scope of this PR or ticket, but could you create a ticket for documenting the assumptions we have for 'flavor' nomenclature so we can have a set of standards to follow going forward?

filariow:
> @filariow this task was already split in two to keep the size of the PR smaller. 

That's not mandatory for us to have 1:1 mapping between tasks and PRs. I'm really struggling reviewing this PR properly. Please split it in multiple ones.

PR #578, opened 2025-10-08
Try to use dynamic ginkgo version in sealights workflow

mshaposhnik:
Hardcoded v2.22 seems to be working wrong way


filariow (changes requested):
we can remove the step all together as we are now using what's in `hack/golang-ci`

PR #846, opened 2026-04-02
Configure Renovate for weekly schedule with release age filter

oswcab:
Change Renovate to run once a week (Monday morning) and require releases to be at least 3 days old before proposing them. This reduces PR noise and avoids picking up broken releases that get yanked shortly after publishing.

Contributes to: KFLUXINFRA-3008

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,5 +1,9 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "timezone": "UTC",
+  "schedule": [
+    "before 7am on Monday"
+  ],
   "prConcurrentLimit": 20,
   "extends": [
     "config:recommended"


.github/renovate.json
```diff
@@ -1,5 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "timezone": "UTC",
+  "schedule": ["before 7am on Monday"],
+  "minimumReleaseAge": "3 days",
```
filariow: is this applied to security fixes too?

filariow (changes requested):

.github/renovate.json
oswcab: Yes, to any update, including security fixes. Issue is, not sure we can tell apart security from non-security one without examining the PR itself. On the other hand, the main goal of this change is to give us some breath while we process all the stale PRs as we haven't been merging them, security impacted or not.

filariow: As I see it they will still pile up, but with 3 days old updates :) this is not reducing the number of  dependency updates we need to review, right?

oswcab: You're right — these changes reduce noise timing and avoid broken releases, but don't reduce volume. The next step it to add dependency grouping (e.g., group all patch updates into one PR) and automerge for patch-level bumps to actually cut down on review burden. Then, the AI assisted evaluation of PRs will allow the engineer in charge of taking faster (hopefully also better) decisions. This is just the first step of that process.

filariow: > these changes reduce noise timing and avoid broken releases

are we suffering from this problem? If that's negligible, I think it would be preferable to have CVE fixes with no 3-days delay instead 😄

oswcab: Removed the 3 days delay.

PR #439, opened 2025-03-31
Tag new VMs and clean up orphaned VMs (AWS & Power)

enkeefe00:
This commit is part of a feature to create a clean up mechanism to delete orphaned VMs (i.e. those associated with a non-existent TaskRun). AWS EC2 VMs and IBM Power Cloud PVMs are tagged with their TaskRun's namespace and name on creation and both cloud providers implement a clean up function that iterates over all VM instances, retrieves the TaskRun tag, and removes any VMs not found to be associated with an existing TaskRun.

PR #848, opened 2026-04-09
Remove unused CleanUpVms code and dead helper methods

mshaposhnik:
Remove the CloudProvider.CleanUpVms interface method and all its implementations (AWS, IBM Power, IBM Z) since it is never invoked anywhere in the codebase. Also remove helper functions whose only production caller was CleanUpVms:

- AWSEc2DynamicConfig.findInstancesWithoutTaskRuns
- IBMPowerDynamicConfig.doesInstanceHaveTaskRun
- IBMZDynamicConfig.findInstancesWithoutTaskRuns
- DynamicHostPool.InstanceTag (unused getter, not part of any interface)
- MockCloud.CleanUpVms and FailCleanUpVMs field from test helpers
 

Made-with: Cursor


glevi-rh:
I don't see much added value with testing it, and not sure why do we need it really, @enkeefe00 you added most of the code in #439, what do you think?

enkeefe00:
I think this was part of a half-completed feature that was never implemented. As long is it isn't being used anywhere, delete it.

glevi-rh (approved):
LGTM

PR #872, opened 2026-04-16
Improve unit test coverage for reconciler/taskrun package

mshaposhnik:
Add tests for DynamicHostPool.Deallocate (33% → 100%), covering all 8 code paths including buildHostPool failure, old/young host termination logic, idle checks, and TerminateInstance errors.

Add tests for DynamicResolver.Allocate (47% → 74%), covering GetState error/FailedState paths, LaunchInstance retry logic, CountInstances failure, and pool-at-capacity waiting state.

Add tests for Local.Allocate and createUserTaskSecret (50% → 100%), covering client.Update failure, secret-already-exists, Create error, and SetOwnerReference failure paths.

Overall package coverage: 74.3% → 81.4%

Made-with: Cursor

### DIFF
--- a/pkg/reconciler/taskrun/dynamicpool_test.go
+++ b/pkg/reconciler/taskrun/dynamicpool_test.go
@@ -28,7 +28,7 @@ var _ = Describe("DynamicHostPool test", func() {
 	BeforeEach(func() {
 		// Initialize the scheme and add the TaskRun type
 		s = runtime.NewScheme()
-		Expect(v1.AddToScheme(s)).To(Succeed())
+		Expect(v1.AddToScheme(s)).Should(Succeed())
 		r = &ReconcileTaskRun{}
 	})
 
@@ -43,8 +43,8 @@ var _ = Describe("DynamicHostPool test", func() {
 			r.client = client
 
 			idle, err := dhp.isHostIdle(r, ctx, "idle-host")
-			Expect(err).NotTo(HaveOccurred())
-			Expect(idle).To(BeTrue())
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(idle).Should(BeTrue())
 		})
 
 		It("should return false if the host is not idle", func(ctx SpecContext) {
@@ -60,8 +60,8 @@ var _ = Describe("DynamicHostPool test", func() {
 			r.client = client
 
 			idle, err := dhp.isHostIdle(r, ctx, selectedHost)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(idle).To(BeFalse())
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(idle).Should(BeFalse())
 		})
 
 		It("should return false if an error occurs", func(ctx SpecContext) {
@@ -81,8 +81,8 @@ var _ = Describe("DynamicHostPool test", func() {
 			idle, err := dhp.isHostIdle(r, ctx, selectedHost)
 
 			// Verify that the error is returned and the host is not considered idle
-			Expect(err).To(MatchError(errToReturn))
-			Expect(idle).To(BeFalse())
+			Expect(err).Should(MatchError(errToReturn))
+			Expect(idle).Should(BeFalse())
 		})
 	})
 
@@ -239,7 +239,7 @@ var _ = Describe("DynamicHostPool test", func() {
 
 			err := dhp.Deallocate(r, ctx, tr, "secret-name", selectedHost)
 
-			Expect(err).To(MatchError(listErr))
+			Expect(err).Should(MatchError(listErr))
 		})
 
 		It("should return error when TerminateInstance fails for old idle selectedHost", func(ctx SpecContext) {
@@ -253,7 +253,7 @@ var _ = Describe("DynamicHostPool test", func() {
 
 			err := dhp.Deallocate(r, ctx, tr, "secret-name", selectedHost)
 
-			Expect(err).To(MatchError(ContainSubstring("terminate failed")))
+			Expect(err).Should(MatchError(ContainSubstring("terminate failed")))
 		})
 	})
 })
--- a/pkg/reconciler/taskrun/dynamicpool_test.go
+++ b/pkg/reconciler/taskrun/dynamicpool_test.go
@@ -4,16 +4,58 @@ package taskrun
 import (
 	"context"
 	"errors"
+	"time"
 
+	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
 	. "github.com/konflux-ci/multi-platform-controller/pkg/constant"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
+
+// deallocTestCloud is a lightweight mock cloud provider for Deallocate tests.
+// It gives fine-grained control over ListInstances/TerminateInstance behaviour
+// without touching the global cloudImpl singleton used by integration tests.
+type deallocTestCloud struct {
+	instances     []cloud.CloudVMInstance
+	failList      bool
+	failTerminate bool
+	terminatedIDs []cloud.InstanceIdentifier
+}
+
+func (m *deallocTestCloud) ListInstances(_ client.Client, _ context.Context, _ string) ([]cloud.CloudVMInstance, error) {
+	if m.failList {
+		return nil, errors.New("list instances failed")
+	}
+	return m.instances, nil
+}
+
+func (m *deallocTestCloud) TerminateInstance(_ client.Client, _ context.Context, id cloud.InstanceIdentifier) error {
+	if m.failTerminate {
+		return errors.New("terminate failed")
+	}
+	m.terminatedIDs = append(m.terminatedIDs, id)
+	return nil
+}
+
+func (m *deallocTestCloud) LaunchInstance(_ client.Client, _ context.Context, _ string, _ string, _ map[string]string) (cloud.InstanceIdentifier, error) {
+	return "", nil
+}
+func (m *deallocTestCloud) GetInstanceAddress(_ client.Client, _ context.Context, _ cloud.InstanceIdentifier) (string, error) {
+	return "", nil
+}
+func (m *deallocTestCloud) CountInstances(_ client.Client, _ context.Context, _ string) (int, error) {
+	return len(m.instances), nil
+}
+func (m *deallocTestCloud) GetState(_ client.Client, _ context.Context, _ cloud.InstanceIdentifier) (cloud.VMState, error) {
+	return cloud.OKState, nil
+}
+func (m *deallocTestCloud) SshUser() string { return "test-user" }
 
 var _ = Describe("DynamicHostPool test", func() {
 	var (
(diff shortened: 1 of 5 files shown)


pkg/reconciler/taskrun/dynamicpool_test.go
```diff
@@ -82,4 +124,190 @@ var _ = Describe("DynamicHostPool test", func() {
...
+			}
+			r.client = fake.NewClientBuilder().WithScheme(s).WithObjects(tr, busyTr).Build()
+
+			err := dhp.Deallocate(r, ctx, tr, "secret-name", selectedHost)
+
+			Expect(err).NotTo(HaveOccurred())
+			for _, id := range mockCloud.terminatedIDs {
+				Expect(string(id)).NotTo(Equal(selectedHost))
+			}
+		})
```
meyrevived: The 3-7 It() blocks can be made DRY with Ginkgo's DescribeTable

```suggestion
DescribeTable("should correctly decide whether to terminate the selectedHost",
		func(ctx SpecContext, selectedHost string, instances []cloud.CloudVMInstance,
			assignTrToHost bool, extraObjects []client.Object, expectTerminated bool) {

			mockCloud.instances = instances

			var objects []client.Object
			if assignTrToHost {
				tr.Labels = map[string]string{AssignedHost: selectedHost}
				objects = append(objects, tr)
			}
			objects = append(objects, extraObjects...)

			r.client =
				fake.NewClientBuilder().WithScheme(s).WithObjects(objects...).Build()

			err := dhp.Deallocate(r, ctx, tr, "secret-name", selectedHost)
			Expect(err).NotTo(HaveOccurred())

			if expectTerminated {
				Expect(mockCloud.terminatedIDs).To(ContainElement(cloud.InstanceIdentifier(selectedHost)))
			} else {
				for _, id := range mockCloud.terminatedIDs {
					Expect(string(id)).NotTo(Equal(selectedHost))
				}
			}
		},
		Entry("young host, no old instances — nothing terminated",
			"young-host",
			[]cloud.CloudVMInstance{
				{InstanceId: "young-host", Address: "1.2.3.4", StartTime: time.Now()},
			},
			false, nil, false,
		),
		Entry("young selectedHost with old secondary host — selectedHost survives",
			"young-host",
			[]cloud.CloudVMInstance{
				{InstanceId: "young-host", Address: "1.2.3.4", StartTime: time.Now()},
				{InstanceId: "old-host", Address: "5.6.7.8", StartTime: time.Now().Add(-1 * time.Hour)},
			},
			false, nil, false,
		),
		// The controller removes the AssignedHost label AFTER Deallocate returns,
		// so the current TaskRun is still labelled and the host appears non-idle.
		Entry("old selectedHost with current TaskRun still assigned — not terminated",
			"old-host",
			[]cloud.CloudVMInstance{
				{InstanceId: "old-host", Address: "1.2.3.4", StartTime: time.Now().Add(-1 * time.Hour)},
			},
			true, nil, false,
		),
		Entry("old host with no TaskRuns referencing it — terminated",
			"old-idle-host",
			[]cloud.CloudVMInstance{
				{InstanceId: "old-idle-host", Address: "1.2.3.4", StartTime: time.Now().Add(-1 * time.Hour)},
			},
			false, nil, true,
		),
		Entry("old selectedHost with other tasks still using it — not terminated",
			"old-busy-host",
			[]cloud.CloudVMInstance{
				{InstanceId: "old-busy-host", Address: "1.2.3.4", StartTime: time.Now().Add(-1 * time.Hour)},
			},
			true,
			[]client.Object{&v1.TaskRun{
				ObjectMeta: metav1.ObjectMeta{
					Name:      "other-task",
					Namespace: "default",
					Labels:    map[string]string{AssignedHost: "old-busy-host"},
				},
			}},
			false,
		),
	)
```

pkg/reconciler/taskrun/dynamicpool_test.go
```diff
@@ -4,17 +4,59 @@ package taskrun
...
+func (m *deallocTestCloud) GetInstanceAddress(_ client.Client, _ context.Context, _ cloud.InstanceIdentifier) (string, error) {
+	return "", nil
+}
+func (m *deallocTestCloud) CountInstances(_ client.Client, _ context.Context, _ string) (int, error) {
+	return len(m.instances), nil
+}
+func (m *deallocTestCloud) GetState(_ client.Client, _ context.Context, _ cloud.InstanceIdentifier) (cloud.VMState, error) {
+	return cloud.OKState, nil
+}
+func (m *deallocTestCloud) SshUser() string { return "test-user" }
```
meyrevived, line 58: I think it the most lightweight mock would be to add `failList`, `failTerminate` and `terminatedIDs` slice to MockCloud and add `failList` check from this mock's `TerminateInstance` to MockCloud's `TerminateInstance`. 
It won't be a huge addition of code glued onto MockCloud just for these tests, and it won't be several sub-mock-clouds that we'll have to remember to update all at the same time if anything has to be changed in the future.

pkg/reconciler/taskrun/provision_dynamic_test.go
meyrevived, line 1: Did you change agents in mid-development of this? dynamic_pool_test.go had a different Ginkgo writing style that this one, and I was about to leave a comment in the general review. I'll leave it now:
Please add the content or at least the coding style standards from [this skill file](https://github.com/meyrevived/infra_superpowers/blob/main/skills/ginkgo-testing-standards/SKILL.md) to your agent and go over these changes. 
Andy and I have been keeping any new testing code added to stick to this standard.

pkg/reconciler/taskrun/provision_dynamic_test.go
```diff
@@ -166,6 +166,122 @@ var _ = Describe("Test Dynamic Host Provisioning", func() {
 		})
 	})
 
+	When("the VM instance enters a failed state while waiting for an address", func() {
+
+		It("should terminate the instance and requeue when GetState returns FailedState", func(ctx SpecContext) {
```
meyrevived, line 171: None of the `Expect() `in this test actually verifies a requeue by checking for a `reconcile.Result`'s `RequeueAfter` field. 
So, either the test description should have "and requeue" removed or also Expect()s something to prove a requeue has been done.

pkg/reconciler/taskrun/provision_dynamic_test.go
```diff
@@ -166,6 +166,122 @@ var _ = Describe("Test Dynamic Host Provisioning", func() {
...
+
+			createUserTaskRun(ctx, client, "test-getstate-err", "linux/arm64")
+			// 1st reconcile: launches instance
+			_, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: userNamespace, Name: "test-getstate-err"}})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// 2nd reconcile: address is empty, GetState fails → requeue (no error returned)
+			result, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: userNamespace, Name: "test-getstate-err"}})
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(result.RequeueAfter).Should(BeNumerically(">", 0))
```
meyrevived: Testing needs to reflect what the code does.

```suggestion
    Expect(result.RequeueAfter).Should(Equal(10 * time.Second))
```

meyrevived (changes requested):
A partial review with what I've seen in the time I had, I'll get back to this later and have a go at the rest of the code

pkg/reconciler/taskrun/provision_dynamic_test.go
mshaposhnik: Can you point to any specific mismatch examples? I don't see much deaviation from the proposed ruleset

pkg/reconciler/taskrun/dynamicpool_test.go
mshaposhnik: I don't get the idea, honestly...
I would prefer to have mocks as simple as possible, ideally representing the signle interface (which is ithis one does) so in case of interface changes it cannot be forgotten as it will not satisfy the interface type anymore

pkg/reconciler/taskrun/provision_dynamic_test.go
mshaposhnik: Fixed

pkg/reconciler/taskrun/provision_dynamic_test.go
mshaposhnik: Added field check

pkg/reconciler/taskrun/dynamicpool_test.go
meyrevived: In case of interface changes you'll have to fix two mocks instead of just one.

pkg/reconciler/taskrun/provision_dynamic_test.go
meyrevived: From the [quick references](https://github.com/meyrevived/infra_superpowers/blob/main/skills/ginkgo-testing-standards/SKILL.md#quick-reference):
1. [To()](https://github.com/mshaposhnik/multi-platform-controller/blob/5d554b842049213c4ef0fd9b7516b8e353ef4b7b/pkg/reconciler/taskrun/dynamicpool_test.go#L86) and [NotTo()](https://github.com/mshaposhnik/multi-platform-controller/blob/5d554b842049213c4ef0fd9b7516b8e353ef4b7b/pkg/reconciler/taskrun/dynamicpool_test.go#L85) instead of [Should()](https://github.com/mshaposhnik/multi-platform-controller/blob/5d554b842049213c4ef0fd9b7516b8e353ef4b7b/pkg/reconciler/taskrun/provision_dynamic_test.go#L44) and [ShouldNot()](https://github.com/mshaposhnik/multi-platform-controller/blob/5d554b842049213c4ef0fd9b7516b8e353ef4b7b/pkg/reconciler/taskrun/provision_dynamic_test.go#L43).
2. [When() blocks gathering several themed It() blocks instead of a group of It()](https://github.com/mshaposhnik/multi-platform-controller/blob/5d554b842049213c4ef0fd9b7516b8e353ef4b7b/pkg/reconciler/taskrun/provision_dynamic_test.go#L36) instead of a bunch of It() blocks "floating" inside the big Describe() block, even when they can be put under the same group - your original It() blocks with the same theme were so similar that I could suggest to put them into a DescribeTable().

Something I don't see in this diff either, or in any of the files is one of the standards Andy brought that I particularly like - [inline function calls inside Expect() lines](https://github.com/meyrevived/infra_superpowers/blob/main/skills/ginkgo-testing-standards/SKILL.md#6-inline-function-calls-in-expect) for compactness, I'll give one example in this file as a change suggestion, but it's only one of many.

All of the files in this PR need to have their style align with the rest of this coding standard.

pkg/reconciler/taskrun/provision_local_test.go
```diff
@@ -54,4 +61,87 @@ var _ = Describe("Test Local Host Provisioning", func() {
...
+
+		It("should return nil when the secret already exists", func(ctx SpecContext) {
+			existing := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{Name: "my-secret", Namespace: "default"},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(existing).Build()
+			r = &ReconcileTaskRun{client: fakeClient, scheme: s}
+
+			err := createUserTaskSecret(r, ctx, tr, "my-secret", map[string][]byte{"host": []byte("localhost")})
+			Expect(err).ShouldNot(HaveOccurred())
```
meyrevived, line 117: An example of [inline function calls in Expect() compact testing style](https://github.com/meyrevived/infra_superpowers/blob/main/skills/ginkgo-testing-standards/SKILL.md#6-inline-function-calls-in-expect):
 
```suggestion
			Expect(createUserTaskSecret(r, ctx, tr, "my-secret", map[string][]byte{"host": []byte("localhost")})).ShouldNot(HaveOccurred())
```

pkg/reconciler/taskrun/provision_dynamic_test.go
```diff
@@ -166,6 +166,123 @@ var _ = Describe("Test Dynamic Host Provisioning", func() {
...
+			// 2nd reconcile: tries again, fails → sets failure count to 2, requeues
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: userNamespace, Name: "test-launch-fail"}})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			tr = getUserTaskRun(ctx, client, "test-launch-fail")
+			Expect(tr.Annotations[CloudFailures]).Should(Equal("2"))
+
+			// 3rd reconcile: retries exceeded (failureCount == 2) → returns error
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: userNamespace, Name: "test-launch-fail"}})
+			Expect(err).Should(HaveOccurred())
```
meyrevived, line 244: Missing error string check

```suggestion
			Expect(err).Should(MatchError(ContainSubstring("launch failed")))
```

pkg/reconciler/taskrun/provision_dynamic_test.go
```diff
@@ -166,6 +166,123 @@ var _ = Describe("Test Dynamic Host Provisioning", func() {
...
+			Expect(result.RequeueAfter).Should(Equal(10 * time.Second))
+
+			// Instance should still exist (not terminated on GetState error)
+			Expect(cloudImpl.Running).Should(Equal(1))
+		})
+	})
+
+	When("launching a new instance fails", func() {
+
+		It("should retry on first launch failure and requeue", func(ctx SpecContext) {
```
meyrevived, line 223: The It() is not just that the first launch fails and is verified - you're running all 3 attempts and then verifies that the 3rd attempt is the last, the title should reflect this.

```suggestion
		It("should exhaust launch  retries then return error"", func(ctx SpecContext) {
```

pkg/reconciler/taskrun/provision_local_test.go
```diff
@@ -54,4 +61,87 @@ var _ = Describe("Test Local Host Provisioning", func() {
...
+			// Use a scheme without TaskRun registered so SetOwnerReference
+			// cannot determine the owner's GVK and returns an error.
+			emptyScheme := runtime.NewScheme()
+			Expect(corev1.AddToScheme(emptyScheme)).To(Succeed())
+
+			fakeClient := fake.NewClientBuilder().WithScheme(emptyScheme).Build()
+			r = &ReconcileTaskRun{client: fakeClient, scheme: emptyScheme}
+
+			err := createUserTaskSecret(r, ctx, tr, "my-secret", map[string][]byte{"host": []byte("localhost")})
+			Expect(err).Should(HaveOccurred())
```
meyrevived, line 144: You set up the test case for a very specific scenario, the Expect() should pin it down to the specific returned error message. Otherwise this test would pass even for a DNS resolution failure, or a nil pointer.

```suggestion
			Expect(err).Should(MatchError(ContainSubstring("no kind is registered")))
```

pkg/reconciler/taskrun/provision_local_test.go
```diff
@@ -54,4 +61,87 @@ var _ = Describe("Test Local Host Provisioning", func() {
...
+			Expect(err).Should(MatchError(updateErr))
+		})
+	})
+
+	Describe("createUserTaskSecret", func() {
+		var (
+			s  *runtime.Scheme
+			r  *ReconcileTaskRun
+			tr *pipelinev1.TaskRun
+		)
```
meyrevived, line 95: There's no real need to declare a variable at the top of a test block if it's not going to be shared across the test cases. If each test case sets an `r` of its own, and not have them set inside the `BeforeEach()` like `s` and `tr`, there's no need for it here. "global" vars like that are usually globally set.

```suggestion
		var (
			s  *runtime.Scheme
			tr *pipelinev1.TaskRun
		)
```

meyrevived (changes requested):
Second and last half of the pull request.
All in all good work, some need for tightening screws, and the coding standard across the entire diff.

pkg/reconciler/taskrun/provision_dynamic_test.go
mshaposhnik: The things you're refering is not added in this PR (but yeah, couple of To()s still sneaked into PR I;ll fix them)

pkg/reconciler/taskrun/dynamicpool_test.go
mshaposhnik: this is not a problem, it's a compile time issue which easy to fix. I prefer this over some messy logic.

pkg/reconciler/taskrun/provision_local_test.go
mshaposhnik: Emm, hm. Not sure if I like it to be 100% mandatory for all possible cases....
What if we polish the rule, and then make it public and then refactor all the code at once in separate refinement PR? (which definitely will happen as we will add more skills/knowlegde instructions for agents)

meyrevived: It's not possible to apply this rule everywhere - for things like functions that return two variables it sometimes works and sometimes not, or to things that return a nil pointer if they actually fail/pass instead of what the Expect is looking for. 

Both Claude and Gemini fell into this trap when I told them to apply this rule on code they produced for me. So, the more there is to refactor, the longer the refactoring effort will be.

pkg/reconciler/taskrun/dynamicpool_test.go
meyrevived: Let me see if I understand:
Doing this once is writing messy logic:

```
type MockCloud struct {
	Running            int
	Terminated         int
	Instances          map[cloud.InstanceIdentifier]MockInstance
	FailGetAddress     bool
	TimeoutGetAddress  bool
	FailGetState       bool
	FailLaunch         bool
	FailCountInstances bool
	failList      bool
	failTerminate bool
	terminatedIDs []cloud.InstanceIdentifier
}

...

func (m *MockCloud) ListInstances(kubeClient runtimeclient.Client, ctx context.Context, instanceTag string) ([]cloud.CloudVMInstance, error) {
    if m.failList {
		return nil, errors.New("list instances failed")
	}

	ret := make([]cloud.CloudVMInstance, 0, len(m.Instances))
	for _, v := range m.Instances {
		ret = append(ret, v.CloudVMInstance)
	}
	return ret, nil
}
...

func (m *MockCloud) TerminateInstance(kubeClient runtimeclient.Client, ctx context.Context, instance cloud.InstanceIdentifier) error {
    if m.failTerminate {
		return errors.New("terminate failed")
	}

	m.Running--
	m.Terminated++
    m.terminatedIDs = append(m.terminatedIDs, id) // or replace Terminated with len(m.terminatedIDs)
	delete(m.Instances, instance)
	return nil
}

```

But fixing compile time issue every time something in the interface changes is better than that?

pkg/reconciler/taskrun/provision_local_test.go
mshaposhnik: Exactly, because to me this rule is pretty much questionable TBH... having 1 line economy vs not having ability to put a debug point on `err` or some more checks.... not yet convincing

pkg/reconciler/taskrun/dynamicpool_test.go
mshaposhnik: OK, finally I'm starting to get your idea...
What you mean is just improving the existing MockCloud from `helpers_test` to be able to return errors..
That's sounds ok, but I will externalize it into a separate file as it is not belongs to particular test anymore.

pkg/reconciler/taskrun/provision_local_test.go
meyrevived: Of course you can debug an error even if its inside a compact single line - just Step Into (F11) instead of Step Over...

pkg/reconciler/taskrun/dynamicpool_test.go
meyrevived: Sure, sounds great

mshaposhnik: this been done, @meyrevived PTAL again

pkg/reconciler/taskrun/provision_local_test.go
mshaposhnik: OK, let's ask @filariow to make the final decision on that. 
I still prefer not to combine the tested method call with result checking (keeping code in Given/When/Then style and not combining the last two into a single line). That gives more clear visibility of what we testing, which result do we expect, and still be able to check returned result imediately without having to "just Step Into (F11) " and run through all the tested function call.

filariow: That's mainly a [cosmetic problem](https://onsi.github.io/gomega/#handling-errors) and we don't enforce a standard on it. To me both approaches are good. As also noted in the [SKILL.md#6-inline-function-calls-in-expect](https://github.com/meyrevived/infra_superpowers/blob/main/skills/ginkgo-testing-standards/SKILL.md#6-inline-function-calls-in-expect) that's not a strict requirement.

```
When NOT to inline:
    Setup is complex
    You need the result for multiple assertions
    Readability suffers
```

Personally, what I find more readable is:
```suggestion
			data := map[string][]byte{"host": []byte("localhost")}
			Expect(createUserTaskSecret(r, ctx, tr, "my-secret", data)).To(Succeed())
```

That's said, on this specific topic, I think the PR owner can accept the cosmetic suggestion from others or discard it.

So @mshaposhnik feel free to pick what you think is more readable for the team; I'd vote for what I suggested above 😄

meyrevived: Great suggestion @filariow - the data coming into the function is much more visible, the function we're here to test is expected to succeed instead of not-fail and the function call is inline with its assertion.

meyrevived (approved):
/lgtm

PR #878, opened 2026-04-22
Improve AWS package unit test coverage

mshaposhnik:
## Summary

- Introduce an `ec2API` interface following the [AWS SDK for Go v2 recommended testing pattern](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/unit-testing.html) to enable mocking of EC2 API calls without external dependencies
- Add unit tests for all CloudProvider methods (`LaunchInstance`, `CountInstances`, `GetInstanceAddress`, `TerminateInstance`, `ListInstances`, `GetState`) and for `SecretCredentialsProvider.Retrieve`
- Covers happy paths, API errors, spot-fallback retry, invalid TaskRun IDs, instance state mapping, and Kubernetes secret retrieval
- Reorganize test files: `ec2_api.go` (interface), `ec2_mock_test.go` (mock types), `ec2_test.go` (ec2.go tests), `ec2_helpers_test.go` (helpers tests)

## Test plan

- [x] `go test ./pkg/aws/` passes (59 specs)
- [x] Coverage: 34.0% → 81.4%
- [x] `go build ./...` clean
- [x] No production behavior changes (the `ec2Client` field defaults to nil; `getEC2Client` creates a real client when nil)


pkg/aws/ec2_helpers_test.go
```diff
@@ -325,6 +329,74 @@ var _ = Describe("AWS EC2 Helper Functions", func() {
...
+				}
+				fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(secret).Build()
+				provider := SecretCredentialsProvider{Name: "aws-creds", Namespace: "ns", Client: fakeClient}
+
+				creds, err := provider.Retrieve(ctx)
+
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(creds.AccessKeyID).Should(Equal("AKID"))
+				Expect(creds.SecretAccessKey).Should(Equal("SECRET"))
+				Expect(creds.SessionToken).Should(Equal("TOKEN"))
```
meyrevived, line 349: If we're expecting to drill-down on three different variables in something, there's no need to also make sure the function that returns it didn't error, or we wouldn't have any of the following variables to assert. 
Our CI doesn't run Ginkgo with the flag for stopping at the first failure, because we want the complete picture of everything that fails and everything that works. Adding a line asserting `provider.Retrieve` didn't fail isn't saving any space in the logs or making the test run time shorter.

```suggestion
				Expect(creds.AccessKeyID).Should(Equal("AKID"))
				Expect(creds.SecretAccessKey).Should(Equal("SECRET"))
				Expect(creds.SessionToken).Should(Equal("TOKEN"))
```

pkg/aws/ec2_mock_test.go
meyrevived, line 1: I wonder if we shouldn't move this, the IBM cloud mock from the next PR and [MockInstance](https://github.com/konflux-ci/multi-platform-controller/pull/872/changes#diff-3edc8ef01def3d6cece0df7c3a5e6c5070faada078c97c85c81944a119e5e1f0) to [test/utils](https://github.com/konflux-ci/multi-platform-controller/tree/main/test/utils)/mocks

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,434 @@
...
+		BeforeEach(func() {
+			mock = &mockEC2Client{}
+			cfg = AWSEc2DynamicConfig{
+				InstanceType: "t4g.medium",
+				KeyName:      "test-key",
+				Ami:          "ami-123",
+				Disk:         40,
+				ec2Client:    mock,
+			}
+		})
```
meyrevived, line 128: These are required for every Describe() block here. When there's a difference in configurations, what changes is not the value of any of the config vals here - it's always something from this example.

Why not have a 

``` go

var _ = Describe("Ec2 Unit Test Suit", func() {
    
    var (
    		mock *mockEC2Client
    		cfg  AWSEc2DynamicConfig
    	)
    
    	JustBeforeEach(func() {
    		mock = &mockEC2Client{}
    		cfg = AWSEc2DynamicConfig{
    			InstanceType: "t4g.medium",
    			KeyName:      "test-key",
    			Ami:          "ami-123",
    			Disk:         40,
    			ec2Client:    mock,
    		}
    	})
```

And save ~9 lines under each Describe() here?

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,434 @@
...
+			It("should return the error immediately", func(ctx SpecContext) {
+				mock.RunInstancesErr = errors.New("ec2 launch failed")
+
+				_, err := cfg.LaunchInstance(nil, ctx, "ns:task", "tag", map[string]string{})
+
+				Expect(err).Should(MatchError(ContainSubstring("failed to launch EC2 instance")))
+			})
+		})
+
+		When("the spot instance launch fails", func() {
```
meyrevived: Asking from curiosity - we're still actually doing spot instances? If not, isn't it dead code we can remove and get slightly better test coverage % also in this way?

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,434 @@
...
+			mock *mockEC2Client
+			cfg  AWSEc2DynamicConfig
+		)
+
+		BeforeEach(func() {
+			mock = &mockEC2Client{}
+			cfg = AWSEc2DynamicConfig{InstanceType: "t4g.medium", ec2Client: mock}
+		})
+
+		When("there are matching running instances", func() {
```
meyrevived: This test is very simplistic for the layers of filtering and verification that CountInstances does.
Before the state and VM type are matched, there's also [a layer of filtering by instance tags, and there need to be two sets of matching tags and values](https://github.com/konflux-ci/multi-platform-controller/blob/c5649b969d93aee91f974f6efea49987c78f8d24/pkg/aws/ec2.go#L149-L150). 
These two lines aren't covered by the test code.

Instead of trying to cram everything into one test case, I suggest you turn this into a DescribeTable where the repeating function has lines 208-210 here and an instance with all of the values and tags and states that will make it be selected, and each Entry() will add a positive instance (an identical instance) or a negative instance (like what you had here, but with good tags, or with bad tags and everything else good etc.)

meyrevived (changes requested):
I ran out of time for now, but there's plenty of comments to work on for now.

The coding standard is better stuck to here, excellent!

pkg/aws/ec2_helpers_test.go
mshaposhnik: What if poorly written code return both result and an err? Imo this check is in place.

pkg/aws/ec2_test.go
mshaposhnik: ok, but claude proposed a slighlty different approach - just wrap into another Describe() , to avoid adding unnecessary complexity with `JustBeforeEach`

pkg/aws/ec2_mock_test.go
mshaposhnik: we'll see with IBM, its possible that iw will require more than one layer or mocking

pkg/aws/ec2_test.go
mshaposhnik: agree, very doubtful that we gonna use this..

pkg/aws/ec2_test.go
mshaposhnik: slightly refactored, but wthat you asking is cannot be mocked - tags filtering happens on AWS side by applying request filters, so adding wrong-tagged instances to the mock response would produce misleading tests that pass for the wrong reason.

pkg/aws/ec2_helpers_test.go
meyrevived: Then writing the test helped you find a bug and keeping the test help you not write code poorly in the future. Despite what it feels like at the moment, this is the reason for writing tests.

pkg/aws/ec2_mock_test.go
meyrevived: All the more reason to keep code for product-functionality in one place and code for testing-functionality in another.

pkg/aws/ec2_test.go
meyrevived: Clever Claude

pkg/aws/ec2_test.go
```diff
@@ -170,24 +170,53 @@ var _ = Describe("Ec2 Unit Test Suit", func() {
...
+			DescribeTable("tag filters and instance filtering",
+				func(ctx SpecContext, extraInstance types.Instance, expectedCount int) {
+					positiveInstance := types.Instance{
+						InstanceId:   aws.String("i-positive"),
+						State:        &types.InstanceState{Name: types.InstanceStateNameRunning},
+						InstanceType: "t4g.medium",
+					}
+
+					instances := []types.Instance{positiveInstance}
+					if extraInstance.InstanceId != nil {
```
meyrevived: When would something like this happen? It's a test, we control every part of the input, why would we need to verify the data can be used in testing?

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+					state, err := cfg.GetState(nil, ctx, "i-123")
+
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(state).Should(Equal(cloud.FailedState))
+				})
+			})
+		})
+
+		Describe("ListInstances", func() {
```
meyrevived, line 337: There's no happy path testing here. Can you add a pingIPAddress mock and hook it up to the `mockEC2Client` so we can have this code covered?

pkg/aws/ec2_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(state).Should(Equal(expectedState))
+				},
+				Entry("running instance should return OKState", types.InstanceStateNameRunning, cloud.OKState),
+				Entry("pending instance should return OKState", types.InstanceStateNamePending, cloud.OKState),
+				Entry("terminated instance should return OKState", types.InstanceStateNameTerminated, cloud.OKState),
+				Entry("stopped instance should return OKState", types.InstanceStateNameStopped, cloud.OKState),
+				Entry("shutting-down instance should return OKState", types.InstanceStateNameShuttingDown, cloud.OKState),
+				Entry("stopping instance should return OKState", types.InstanceStateNameStopping, cloud.OKState),
```
meyrevived: Image

```suggestion
				func(ctx SpecContext, stateName types.InstanceStateName) {
					mock.DescribeInstancesOutput = &ec2.DescribeInstancesOutput{
						Reservations: []types.Reservation{{
							Instances: []types.Instance{{
								State: &types.InstanceState{Name: stateName},
							}},
						}},
					}

					state, err := cfg.GetState(nil, ctx, "i-123")

					Expect(err).ShouldNot(HaveOccurred())
					Expect(state).Should(Equal(cloud.OKState))
				},
				Entry("running instance should return OKState", types.InstanceStateNameRunning),
				Entry("pending instance should return OKState", types.InstanceStateNamePending),
				Entry("terminated instance should return OKState", types.InstanceStateNameTerminated),
				Entry("stopped instance should return OKState", types.InstanceStateNameStopped),
				Entry("shutting-down instance should return OKState", types.InstanceStateNameShuttingDown),
				Entry("stopping instance should return OKState", types.InstanceStateNameStopping),
```

meyrevived (changes requested):
Last part of the review. 
Sorry I didn't finish it yesterday as I said I would - I almost fell asleep on the keyboard.

pkg/aws/ec2_test.go
mshaposhnik: Yeah, I saw this too. But there;s no way to mock the package function witout intoducing quite a mess trying to replace the real call with mock call :- (

meyrevived: Have the mock pingIPAddress ping 8.8.8.8, (or 192.0.2.42 for testing failures)

mshaposhnik: I seriously doubt Google will allow us to ping their SSH port.

meyrevived: Had a second look at pingIPAddress, it only returns an error if there was one. Easiest mock in the world.

mshaposhnik: Can you provide an code proposal so I can get your idea of mocking better?

pkg/aws/ec2_helpers_test.go
mshaposhnik: I don't get it... is it agree or disagree?

pkg/aws/ec2_mock_test.go
mshaposhnik: I rephrase my answer - can we leave it so far as-is until IBM PR, because I don't know how complex IBM mocks going to be, and if they might have dependency on something else in the ibm package, it might introduce the circular depenency.

meyrevived: Sure, that makes sense

pkg/aws/ec2_test.go
meyrevived: You came up with a great way to do it yourself [here](https://github.com/konflux-ci/multi-platform-controller/pull/878/commits/35942f0fecbe6b3b95965fbbb801c91b8d45f1c7), you didn't need me

meyrevived (approved):
/lgtm

PR #883, opened 2026-04-30
Improve IBM Power package unit test coverage

mshaposhnik:
## Summary

- Introduce a `powerAPI` interface to decouple CloudProvider methods from the IBM Power REST client, enabling unit tests with a hand-written mock (same pattern as the AWS `ec2API`/`mockEC2Client` approach from #878).
- Refactor `ibmp_helpers.go` by moving API methods (`listInstances`, `launchInstance`, `getInstance`, `deleteInstance`, `resizeInstanceVolume`, `updateVolume`) into a dedicated `ibmp_client.go` behind the new interface, leaving only true helpers (`retrieveInstanceIp`, `parseCRN`, `createAuthenticatedBaseService`).
- Add injectable `pClient` and `pingFunc` fields to `IBMPowerDynamicConfig` to bypass real IBM API and network calls in tests.
- Add 20 new test cases covering all CloudProvider methods: `LaunchInstance`, `CountInstances`, `GetInstanceAddress`, `ListInstances`, `GetState`, and `TerminateInstance`.

**Coverage for `pkg/ibm` rises from 8.0% to 27.4%**, with Power CloudProvider methods going from 0% to 85-93%.

## Test plan

- [x] All 64 existing + new tests pass (`go test ./pkg/ibm/ -v`)
- [x] No linter errors introduced
- [x] Package compiles cleanly
- [x] Coverage verified before/after via `go test -coverprofile`


pkg/ibm/ibmp_test.go
```diff
@@ -1,22 +1,13 @@
...
-//	- Configuration data is passed to IBMPowerDynamicConfig correctly when the values are valid
-//  - Default values are inserted whenever the configuration written to host-config.yaml are problematic in structure or value
-//
-// There are 5 test cases:
-// 	1. A positive test to verify all is working correctly with valid config map keys
-//	2. A negative test with a platform name unlike any the MPC covers
-//	3. A negative test to verify default value completion - empty memory, core number and disk size values
-//	4. A negative test to verify default value completion - non-numeric memory, core number and disk size values
-//	5. A negative test to verify default value completion - Verifying disk size default number of 100 if the configuration aims for less than that
-// Assisted-by: TAG
```
meyrevived, line 12: Since the test for CreateIBMPowerCloudConfig is still here in its original form, just nested in the bigger test suite, the TAG-assisted code is still here.

You could move the description of CreateIBMPowerCloudConfig's test just above its block, and add the name of the agent you used to generate this after TAG

pkg/ibm/ibmp.go
meyrevived, line 194: Since you've done some refactoring on Emily/Stuart's code when you moved all of the helper code to this file, I'm taking the liberty to also add a fix for something that jumped at me:

If we have a machine running for over a year, we better recycle it, but not immediately fail a TaskRun over it (or we can fail it, if we want users to be actually angry at _us_ because a TR for IBM failed).
Also, if something got corrupted in the REST API response and Go tried to interpret the garbage as 711AD, or if the corruption means we get an empty field in CreationDate and Go says the machine was created in January 1 1970 - we're covered.

I'm also adding a test case for this so you won't get barked at by Codecov

```suggestion
		createdAt := time.Time(instance.CreationDate)
		
	   if time.Since(createdAt) > 365*24*time.Hour {                                                                                                               
           log.Info("WARN: instance creation date is over one year old — possible data integrity issue",
           "instanceID", identifier,                                                                                                                           
            "creationDate", createdAt,
      )                                                                                                                                                       
  }
```

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +22,424 @@ func encodeUserData(s string) string {
...
+						},
+					}
+
+					instances, err := cfg.ListInstances(nil, ctx, "my-tag")
+
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(instances).Should(BeEmpty())
+				})
+			})
+		})
```
meyrevived, line 486: Here's the test for the CreationDate validation code

```suggestion
			})
			
		When("an instance has a suspiciously old creation date", func() {                                                                                           
          DescribeTable("it should still include the instance in the results",
              func(ctx SpecContext, creationDate strfmt.DateTime) {
                  mock.ListInstancesOutput = models.PVMInstances{                                                                                                 
                      PvmInstances: []*models.PVMInstanceReference{
                          {                                                                                                                                       
                              ServerName:    ptr("moshe-kipod-tag-123x"),
                              PvmInstanceID: ptr("id-1"),                                                                                                         
                              Networks:      []*models.PVMInstanceNetwork{{ExternalIP: "1.2.3.4"}},
                              CreationDate:  creationDate,                                                                                                        
                          },                                                                                                                                      
                      },
                  }                                                                                                                                               
                  
                  instances, err := cfg.ListInstances(nil, ctx, "my-tag")                                                                                         
  
                  Expect(err).ShouldNot(HaveOccurred())                                                                                                           
                  Expect(instances).Should(HaveLen(1))
                  Expect(string(instances[0].InstanceId)).Should(Equal("id-1"))                                                                                   
              },                                                                                                                                                  
              Entry("when CreationDate is empty (null or missing from API)", 
              strfmt.DateTime{},),                                                                                                                                                                                                                                                                                                                                                                   
              Entry("when CreationDate is corrupt data interpreted as Unix epoch",
              strfmt.DateTime(time.Unix(0, 0)),                                                                         
              ),                                                                                                                                                  
          )           
      })
	})
```

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,271 @@
+package ibm
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
```
meyrevived, line 9: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part I

```suggestion
	"fmt"
	"net/http"
	"strings"
```

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	requestBuilder.AddHeader("CRN", pc.config.CRN)
+	requestBuilder.AddHeader("Accept", "application/json")
+	request, err := requestBuilder.Build()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build the HTTP request: %w", err)
+	}
+
+	instance := models.PVMInstance{}
```
meyrevived, line 160: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part III

```suggestion
	}
	
	reqFn := pc.requestFn
	if reqFn == nil {
		reqFn = pc.service.Request
	}

	instance := models.PVMInstance{}
```

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	requestBuilder.AddHeader("CRN", pc.config.CRN)
+	requestBuilder.AddHeader("Content-Type", "application/json")
+	requestBuilder.AddHeader("Accept", "application/json")
+
+	request, err := requestBuilder.Build()
+	if err != nil {
+		return fmt.Errorf("failed to build the HTTP request: %w", err)
+	}
+	var vRef models.VolumeReference
```
meyrevived, line 285: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part IV

```suggestion
	}
	
	reqFn := pc.requestFn
	if reqFn == nil {
		reqFn = pc.service.Request
	}
	
	var vRef models.VolumeReference
```

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/go-logr/logr"
+	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
+)
+
+// powerClient implements powerAPI using the IBM Power Virtual Server REST API
+// via core.BaseService.
+type powerClient struct {
+	service *core.BaseService
+	config  IBMPowerDynamicConfig
+}
```
meyrevived, line 32: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part II

```suggestion
	config  IBMPowerDynamicConfig
	requestFn func(req *http.Request, result interface{}) (*core.DetailedResponse, error) // requestFn allows tests to inject a mock for pc.service.Request calls.
}
```

pkg/ibm/ibmp_test.go
```diff
@@ -1,22 +1,13 @@
...
 package ibm
 
 import (
+	"context"
 	"encoding/base64"
+	"errors"
 	"strconv"
 
+	"github.com/IBM-Cloud/power-go-client/power/models"
+	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
```
meyrevived, line 15: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part V

```suggestion
	"errors"
	"fmt"
	"net/http"
    "strconv"

	"github.com/IBM-Cloud/power-go-client/power/models"
	"github.com/IBM/go-sdk-core/v5/core"
	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
```

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +22,424 @@ func encodeUserData(s string) string {
...
+			mock *mockPowerClient
+			cfg  IBMPowerDynamicConfig
+		)
+
+		BeforeEach(func() {
+			mock = &mockPowerClient{}
+			cfg = IBMPowerDynamicConfig{
+				pClient:  mock,
+				pingFunc: func(_ context.Context, _ string) error { return nil },
+			}
```
meyrevived: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part VI


```suggestion
    service, err := core.NewBaseService(&core.ServiceOptions{
		URL:           "https://test.power-iaas.cloud.ibm.com",
		Authenticator: &core.NoAuthAuthenticator{},
	})
	Expect(err).ShouldNot(HaveOccurred())

	cfg = IBMPowerDynamicConfig{
		CRN:      "crn:v1:bluemix:public:power-iaas:dal10:a/123456789:test-service-id::",
		Url:      "https://test.power-iaas.cloud.ibm.com",
		Disk:     200,
		pingFunc: func(_ context.Context, _ string) error { return nil },
	}

	mock = &mockPowerClient{
		powerClient: powerClient{
		    service: service,
			config:  cfg,
		},
	}

	cfg.pClient = mock
```

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +22,424 @@ func encodeUserData(s string) string {
...
+
+			When("the initial delete fails", func() {
+				It("should still return nil (error is swallowed, retry loop continues)", func(ctx SpecContext) {
+					mock.DeleteInstanceErr = errors.New("delete failed")
+
+					Expect(cfg.TerminateInstance(nil, ctx, "pvm-123")).ShouldNot(HaveOccurred())
+				})
+			})
+		})
+	})
```
meyrevived, line 614: Changes for suggested requestFn for adding coverage for updateInstanceVolume and resizeIntanceVolume the same clever way as pingFn was used - part VII and last

```suggestion
		})
		
		Describe("updateVolume", func() {
			When("the API request succeeds", func() {
				It("should update the volume without error", func(ctx SpecContext) {
					volumeID := "test-vol-id"
					mock.requestFn = func(req *http.Request, result interface{}) (*core.DetailedResponse, error) {
						Expect(req.Method).To(Equal("PUT"))
						Expect(req.URL.Path).To(ContainSubstring(volumeID))

						vRef, ok := result.(*models.VolumeReference)
						Expect(ok).To(BeTrue())
						vRef.VolumeID = &volumeID
						size := 200.0
						vRef.Size = &size
						return &core.DetailedResponse{StatusCode: 200}, nil
					}

					Expect(mock.updateVolume(ctx, volumeID)).ShouldNot(HaveOccurred())
				})
			})

			When("the API request fails", func() {
				It("should return a descriptive error", func(ctx SpecContext) {
					mock.requestFn = func(_ *http.Request, _ interface{}) (*core.DetailedResponse, error) {
						return nil, fmt.Errorf("API timeout")
					}

					Expect(mock.updateVolume(ctx, "vol-123")).
						Should(MatchError(ContainSubstring("failed to update volume")))
				})
			})

			When("the CRN is invalid", func() {
				It("should return a CRN parse error", func(ctx SpecContext) {
					mock.config.CRN = "not-a-crn"

					Expect(mock.updateVolume(ctx, "vol-123")).
						Should(MatchError(ContainSubstring("failed to retrieve cloud service instance ID")))
				})
			})
		})

		Describe("resizeInstanceVolume", func() {
			When("the volume size differs from the target", func() {
				It("should call updateVolume to resize", func(ctx SpecContext) {
					instanceID := "test-instance"
					volumeID := "test-volume"
					currentDisk := 100.0
					targetDisk := 200.0
					resizeDone := make(chan struct{})

					mock.config.Disk = targetDisk
					mock.requestFn = func(req *http.Request, result interface{}) (*core.DetailedResponse, error) {
						switch req.Method {
						case "GET":
							inst, ok := result.(*models.PVMInstance)
							Expect(ok).To(BeTrue())
							inst.PvmInstanceID = &instanceID
							inst.VolumeIDs = []string{volumeID}
							inst.DiskSize = &currentDisk
							return &core.DetailedResponse{StatusCode: 200}, nil
						case "PUT":
							vRef, ok := result.(*models.VolumeReference)
							Expect(ok).To(BeTrue())
							vRef.VolumeID = &volumeID
							vRef.Size = &targetDisk
							close(resizeDone)
							return &core.DetailedResponse{StatusCode: 200}, nil
						}
						return nil, fmt.Errorf("unexpected method: %s", req.Method)
					}

					mock.resizeInstanceVolume(ctx, &instanceID)

					Eventually(resizeDone).WithTimeout(15 * time.Second).Should(BeClosed())
				})
			})

			When("the volume size already matches the target", func() {
				It("should not call updateVolume", func(ctx SpecContext) {
					instanceID := "test-instance"
					diskSize := 200.0
					updateCalled := make(chan struct{})
					getDone := make(chan struct{}, 1)

					mock.config.Disk = diskSize
					mock.requestFn = func(req *http.Request, result interface{}) (*core.DetailedResponse, error) {
						if req.Method == "GET" {
							inst, ok := result.(*models.PVMInstance)
							Expect(ok).To(BeTrue())
							inst.PvmInstanceID = &instanceID
							inst.VolumeIDs = []string{"vol-1"}
							inst.DiskSize = &diskSize
							select {
							case getDone <- struct{}{}:
							default:
							}
							return &core.DetailedResponse{StatusCode: 200}, nil
						}
						close(updateCalled)
						return nil, nil
					}

					mock.resizeInstanceVolume(ctx, &instanceID)

					Eventually(getDone).WithTimeout(15 * time.Second).Should(Receive())
					Consistently(updateCalled).WithTimeout(1 * time.Second).ShouldNot(BeClosed())
				})
			})
		})
		
	})
```

meyrevived (changes requested):
Looks like ppcle needed some squirming, but you found a way to mock it, great!

I added a few comments here and there, and also some code blocks for adding coverage using the same clever trick we used to mock pinging in AWS unit tests.

pkg/ibm/ibmp_test.go
mshaposhnik: Assisted-by: TAG
Removed-by: Cursor
🤦

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +25,579 @@ func encodeUserData(s string) string {
...
+					}
+
+					Expect(mock.updateVolume(ctx, volumeID)).ShouldNot(HaveOccurred())
+				})
+			})
+
+			When("the API request fails", func() {
+				It("should return a descriptive error", func(ctx SpecContext) {
+					mock.requestFn = func(_ *http.Request, _ interface{}) (*core.DetailedResponse, error) {
+						return nil, fmt.Errorf("API timeout")
```
meyrevived:
```suggestion
						return nil, errors.New("API timeout")
```

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/go-logr/logr"
+	"github.com/konflux-ci/multi-platform-controller/pkg/cloud"
+)
+
+// powerClient implements powerAPI using the IBM Power Virtual Server REST API
+// via core.BaseService.
+type powerClient struct {
+	service *core.BaseService
+	config  IBMPowerDynamicConfig
+	config  IBMPowerDynamicConfig
```
meyrevived: Remove duplicated line
```suggestion

```

meyrevived (dismissed):
Fixing what's been broken

pkg/ibm/ibmp_client.go
mshaposhnik: @meyrevived something is not adding up.. could you check if you share the changes in mock itself?

pkg/ibm/ibmp_client.go
mshaposhnik: Can you check if you must also commit the mock changes?
I cannot convince Cursor to do so, it says profit does not worth the hussle:

```
Not necessarily a bad idea, but it would be mixing two different 
abstraction levels in the same mock, which gets confusing:
mockPowerClient exists to mock the powerAPI interface — 
the 4 operations that CloudProvider methods call. It's clean and focused.

requestFn, updateVolume, and resizeInstanceVolume are internals 
of the real powerClient (HTTP plumbing). They're not part of powerAPI.
If you bolt them onto mockPowerClient, the mock becomes 
a hybrid — part interface mock, part real implementation test 
fixture. Anyone reading the tests later would wonder 
"wait, which powerClient am I testing?"
The real question is: is testing updateVolume/resizeInstanceVolume 
worth the effort?
Those tests were trying to verify that the powerClient correctly 
builds HTTP requests and handles responses. That's the exact 
plumbing layer we intentionally put behind the interface. 
The logic there is mostly mechanical (build URL, set headers, 
call service.Request, unmarshal). 
And resizeInstanceVolume has a goroutine + sleep loop, making it flaky to test.
```

meyrevived: Tell your Cursor that my Claude said:

```
Honest take: Cursor raises a fair point about the abstraction mixing, but I think it's arguing against a strawman version of what these tests actually do.
                                                                                      
  What the tests actually verify:                                                                                                                                            
                                                                                                                                                                             
  - updateVolume: Parses CRN, builds the correct URL with cloud/volume path params, sets the right headers, marshals the UpdateVolume body, and handles errors. Yes, that's  
  "mechanical," but a wrong CRN parse or a misplaced path param silently sends requests to the wrong endpoint. Those are real bugs that code review won't reliably catch.    
  - resizeInstanceVolume: This is the stronger case. It has branching logic — compare disk sizes, skip resize if they match, retry on failure, respect timeouts. That's not  
  plumbing, that's business logic buried inside a goroutine. The "it's flaky to test" argument is actually an argument for testing it — if the goroutine behavior is hard to 
  reason about, you want a test pinning it down.
                                                                                                                                                                             
  Are we testing the mock or the real code?                                                                                                                                  
  
  Here's the key: updateVolume and resizeInstanceVolume are methods on the real powerClient, not on the mock. The mock just embeds powerClient, so when the test calls       
  mock.updateVolume(), it runs the actual production code. The only thing mocked is the HTTP transport (requestFn), which is exactly the boundary you'd want to mock — like
  how you'd mock an HTTP client in any other language.                                                                                                                       
                  
  Compare: mockPowerClient mocks powerAPI (the 4 interface methods) so CloudProvider tests don't hit IBM's API. Our requestFn mocks core.BaseService.Request so powerClient  
  tests don't hit IBM's API. Same pattern, one layer down. The "two abstraction levels" objection applies equally to any test that uses a test double at a transport
  boundary.                                                                                                                                                                  
                  
  Where Cursor has a point:

  The embedding does make mockPowerClient serve double duty. If someone reads the mock file cold, the powerClient embed isn't self-explanatory. But the alternative — a      
  separate testPowerClient with its own setup — is exactly what we started with, and you correctly pushed back on that as unnecessary duplication since both need the same
  service + config.                                                                                                                                                          
                  
  The bottom line: these tests exercise real production logic (CRN parsing, URL construction, resize-vs-skip branching, goroutine lifecycle) with only the HTTP transport    
  faked out. That's not testing the mock — that's textbook unit testing of code that happens to make HTTP calls.
                                                                                                                                                                             
  The "is it worth the effort" question is really about resizeInstanceVolume's 10-second sleep making the tests slow. That's a valid tradeoff concern, but it's about test   
  speed, not test value.
```

mshaposhnik: cool. can you check if you can make the mock refactoring proposal

meyrevived: What refactoring proposal?

mshaposhnik: extending the mock with the new field + methods, which you call in the tests for update/resize volume

meyrevived: Make the proposal to who? I added the code in the proposal to this PR and you added my suggestions to it. The errors in the CI come from fmt and lint bugs I added and the last two code suggestions I opened are the fix for them.

mshaposhnik: Uff...
Can you checkout the branch, and look at the test fucntions you added? They have compilation errors because calling/setting things which are not exist in mock, like `mock.requestFn()`
That makes me thinking you forgot to send the changing proposal for the mock object, no?

meyrevived: All the issues are fixed now

meyrevived (approved):
/lgtm

pkg/ibm/ibmp.go
```diff
@@ -147,31 +165,42 @@ func (pw IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
...
-	pvmInstancesCollection, err := pw.listInstances(ctx, service)
+	pvmInstancesCollection, err := pwClient.listInstances(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch Power Systems instances: %w", err)
 	}
 
+	pingFn := pw.pingFunc
+	if pingFn == nil {
+		pingFn = checkIfIpIsLive
+	}
```
filariow, line 186: Haven't tested it, but I'd expect to be able to simplify it via `cmp.Or`

```suggestion
	pingFn := cmp.Or(pw.pingFunc, checkIfIpIsLive)
```

pkg/ibm/ibmp.go
```diff
@@ -147,31 +165,42 @@ func (pw IBMPowerDynamicConfig) GetInstanceAddress(kubeClient client.Client, ctx
...
 	vmInstances := make([]cloud.CloudVMInstance, 0, len(pvmInstancesCollection.PvmInstances))
-	// Ensure all listed instances have a reachable IP address
 	for _, instance := range pvmInstancesCollection.PvmInstances {
 		if !strings.HasPrefix(*instance.ServerName, instanceTag) {
 			continue
 		}
 		identifier := cloud.InstanceIdentifier(*instance.PvmInstanceID)
 		createdAt := time.Time(instance.CreationDate)
+
+		if time.Since(createdAt) > 365*24*time.Hour {
```
filariow: I'd suggest to move `365*24*time.Hour` to a const, so it's calculated at compile time

```
const taskRunCreatedSince = 365 * 24 * time.Hour
```

pkg/ibm/ibmp.go
```diff
@@ -298,4 +322,12 @@ type IBMPowerDynamicConfig struct {
...
 	// Possible values are "dedicated", "shared", and "capped".
 	ProcType string
+
+	// pClient allows tests to inject a mock Power API client.
+	// When nil, getPowerClient builds a real client from IBM credentials.
+	pClient powerAPI
+
+	// pingFunc allows tests to inject a mock for SSH connectivity checks.
+	// When nil, the real checkIfIpIsLive (TCP dial to port 22) is used.
+	pingFunc func(ctx context.Context, ip string) error
```
filariow, line 337: I'm not a fan of having unexported fields in production code for tests only, but I can't think of an easy way to avoid this. Maybe turning them in a feature? In other words, exporting them and expecting them to be set 😄 

Just bumping ideas to start a conversation. I don't want to block this PR on this topic. Feel free to resolve this comment, move the conversation elsewhere, and merge the PR if needed.

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,284 @@
...
+	pathParamsMap := map[string]string{
+		"cloud": cloudId,
+	}
+	_, err = requestBuilder.ResolveRequestURL(pc.config.Url, `/pcloud/v1/cloud-instances/{cloud}/pvm-instances`, pathParamsMap)
+	if err != nil {
+		return models.PVMInstances{}, fmt.Errorf("failed to encode the request with parameters: %w", err)
+	}
+
+	requestBuilder.AddHeader("CRN", pc.config.CRN)
+	requestBuilder.AddHeader("Accept", "application/json")
```
filariow: minor: we could use `const`s to store these value to avoid repeating them

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,284 @@
...
+	requestBuilder = requestBuilder.WithContext(ctx)
+	requestBuilder.EnableGzipCompression = pc.service.GetEnableGzipCompression()
+
+	cloudId, err := pc.config.parseCRN()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cloud service instance ID: %w", err)
+	}
+	pathParamsMap := map[string]string{
+		"cloud":           cloudId,
+		"pvm_instance_id": pvmId,
```
filariow, line 179: minor: we can use `const` for these keys to avoid repeating them

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return err
+}
+
+// resizeInstanceVolume asynchronously resizes the id instance's volume.
+func (pc *powerClient) resizeInstanceVolume(ctx context.Context, id *string) {
+	log := logr.FromContextOrDiscard(ctx)
+	sleepTime := 10
+	timeout := time.Now().Add(time.Minute * 10)
+	go func() {
```
filariow, line 207: this is a fire and forget approach, which is usually considered unsafe. Can we improve it? 

I see this is not NEW code. Feel free to track this in a JIRA Task and punt to a follow up PR

pkg/ibm/ibmp_client.go
```diff
@@ -0,0 +1,284 @@
...
+		err = fmt.Errorf("failed to delete the Power System VM instance %s: %w", pvmId, err)
+	}
+	return err
+}
+
+// resizeInstanceVolume asynchronously resizes the id instance's volume.
+func (pc *powerClient) resizeInstanceVolume(ctx context.Context, id *string) {
+	log := logr.FromContextOrDiscard(ctx)
+	sleepTime := 10
+	timeout := time.Now().Add(time.Minute * 10)
```
filariow, line 206: instead of counting time ourselves we can craft a context with timeout and check in the `for` loop if the context is still valid. Something like

```go
	tctx := context.WithTimeout(ctx, 10*time.Minute)
	for {
		if err := tctx.Err(); err != nil {
			return err
		}

		// my logic

		time.Sleep(sleepTime)
	}

```

or else we can take advantage of similar logic already provided by `k8s.io/apimachinery/pkg/util/wait`, like `UntilWithContext`

I see this is not NEW code. Feel free to track this in a JIRA Task and punt to a follow up PR

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +27,592 @@ func encodeUserData(s string) string {
...
+				providerConfig := provider.(IBMPowerDynamicConfig)
+				Expect(providerConfig).ToNot(BeNil())
+
+				Expect(providerConfig.Key).Should(Equal("test-key"))
+				Expect(providerConfig.ImageId).Should(Equal("test-image"))
+				Expect(providerConfig.Secret).Should(Equal("test-secret"))
+				Expect(providerConfig.Url).Should(Equal("test-url"))
+				Expect(providerConfig.CRN).Should(Equal("test-crn"))
+				Expect(providerConfig.Network).Should(Equal("test-network"))
+				Expect(providerConfig.System).Should(Equal("test-system"))
```
filariow, line 71: minor: if we expect them to be the same as the ones provided in `config` we can match against those values instead.


```go
				Expect(providerConfig.Key).Should(Equal(config["dynamic." + platform + ".key"]))
				...
```

pkg/ibm/ibmp_test.go
```diff
@@ -31,66 +27,592 @@ func encodeUserData(s string) string {
...
+				"userData": commonUserData,
+				"memory":   "64.0",
+				"cores":    "8.0",
+				"disk":     "42"}, commonUserData, "64.0", "8.0", "100"),
+		)
+	})
+
+	Describe("SshUser", func() {
+		It("should return root", func() {
+			Expect(IBMPowerDynamicConfig{}.SshUser()).Should(Equal("root"))
```
filariow: do we have a `const` for "root"?

pkg/ibm/ibmp.go
mshaposhnik: `cmp.Or` wants `comparable` arguments, which func-s are not

pkg/ibm/ibmp_test.go
mshaposhnik: we will

pkg/ibm/ibmp_client.go
mshaposhnik: Moved everything that repeats

pkg/ibm/ibmp_client.go
mshaposhnik: Moved everything that repeats

pkg/ibm/ibmp_client.go
mshaposhnik: We'll do it  separately outside of testing PR

pkg/ibm/ibmp_client.go
mshaposhnik: We'll do it  separately outside of testing PR

pkg/ibm/ibmp_test.go
mshaposhnik: That's a `testing go assinment` vs `testing go assigment + actual values`  question, so I'll better leave it as-is so at least it pretends to be testing something more than just `=` operation :laughing:

pkg/ibm/ibmp.go
meyrevived: When Max is done with s390x coverage enhancements, I want to move all of the mocking code to the tests/utils/ dir in MPC. We can have a file with mock utils like pingFunc and requestFunc there and still keep the use of a pingFunc inside this code. Just an idea, not actually Go-backed.

pkg/ibm/ibmp_test.go
meyrevived: This is a test that was here before Max's PR. I wrote it with TAG at the time, three ice ages ago. Technically not relevant to this PR.
But you can always create a code change suggestion and Max can commit it 😁

pkg/ibm/ibmp.go
mshaposhnik: Valid point, that;s the biggest challenge of testing the foreign APIs - to substitute them..
if you don't mind I'll keep it as is for the time being to keep same approach with AWS IBM drivers..
If we have better ideas meanwile, we;ll apply them.

pkg/ibm/ibmp_client.go
mshaposhnik: I've created https://redhat.atlassian.net/browse/KFLUXINFRA-3823 out of those two comments

filariow (approved):

PR #933, opened 2026-06-11
KFLUXINFRA-1234 Make otelcol-contrib installation non-fatal when GitHub is unavailable

mshaposhnik:
## Summary

- Added `--fail` to `curl` so that HTTP error responses (e.g. 504 Gateway Timeout HTML bodies) are detected as failures rather than silently saved as a corrupt file.
- Wrapped the download and RPM install in a non-fatal `if/elif/else` block: when GitHub is unreachable, a warning is printed and the otelcol step is skipped entirely instead of aborting provisioning.
- Bumped `otelcol-contrib` version from `0.140.0` to `0.154.0` (latest release as of 2026-06-09).

The host still provisions successfully (users, SSH keys, etc.) when GitHub is down. On the next provisioning run for that host, otelcol will be retried automatically because the binary won't be present.

## Test plan

- [ ] Verify provisioning succeeds end-to-end when GitHub is reachable (happy path, otelcol installed as before).
- [ ] Simulate GitHub unavailability (e.g. bad URL or network block) and confirm the script prints the warning and continues past the otelcol block without exiting.

## Related issues

Fixes the 504-caused provisioning failure reported in shared host provisioning logs.

### DIFF
--- a/deploy/operator/provision-shared-host.sh
+++ b/deploy/operator/provision-shared-host.sh
@@ -40,27 +40,31 @@ if command -v otelcol-contrib >/dev/null 2>&1; then
   sudo systemctl restart otelcol-contrib
 else
   if [[ -f /etc/otelcol-contrib/config_mpc.yaml ]]; then
-    PKG="otelcol-contrib_0.140.0_${PLATFORM}.rpm"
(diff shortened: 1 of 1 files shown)


mshaposhnik:
Reopening from upstream branch so e2e tests can run.

PR #934, opened 2026-06-11
KFLUXINFRA-3956 Make otelcol-contrib installation non-fatal when GitHub is unavailable

mshaposhnik:
## Summary

- Added `--fail` to `curl` so that HTTP error responses (e.g. 504 Gateway Timeout HTML bodies) are detected as failures rather than silently saved as a corrupt file.
- Wrapped the download and RPM install in a non-fatal `if/elif/else` block: when GitHub is unreachable, a warning is printed and the otelcol step is skipped entirely instead of aborting provisioning.
- Bumped `otelcol-contrib` version from `0.140.0` to `0.151.0` 

The host still provisions successfully (users, SSH keys, etc.) when GitHub is down. On the next provisioning run for that host, otelcol will be retried automatically because the binary won't be present.

## Test plan

- [ ] Verify provisioning succeeds end-to-end when GitHub is reachable (happy path, otelcol installed as before).
- [ ] Simulate GitHub unavailability (e.g. bad URL or network block) and confirm the script prints the warning and continues past the otelcol block without exiting.

## Related issues

Fixes the 504-caused provisioning failure reported in shared host provisioning logs.

### DIFF
--- a/deploy/operator/provision-shared-host.sh
+++ b/deploy/operator/provision-shared-host.sh
@@ -40,27 +40,31 @@ if command -v otelcol-contrib >/dev/null 2>&1; then
   sudo systemctl restart otelcol-contrib
 else
   if [[ -f /etc/otelcol-contrib/config_mpc.yaml ]]; then
-    PKG="otelcol-contrib_0.140.0_${PLATFORM}.rpm"
-    URL="https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.140.0/\$PKG"
+    PKG="otelcol-contrib_0.154.0_${PLATFORM}.rpm"
+    URL="https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.154.0/\$PKG"
     echo "Downloading: \$URL"
-    curl -LO "\$URL" || exit 1
-    sudo rpm -ivh "\$PKG" || exit 1
-    # Ensure config dir exists
-    sudo install -d /etc/otelcol-contrib
+    if ! curl -LO --fail "\$URL"; then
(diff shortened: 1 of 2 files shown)


mshaposhnik:
We can skip Qodo warning about not having updates - for dynamic machines it is not applicable, an statics are updated on demand by update script

oswcab:
Only comment is we better squash these commits when merging so history is a bit cleaner

deploy/operator/provision-shared-host.sh
```diff
@@ -40,27 +40,31 @@ if command -v otelcol-contrib >/dev/null 2>&1; then
   sudo systemctl restart otelcol-contrib
 else
   if [[ -f /etc/otelcol-contrib/config_mpc.yaml ]]; then
-    PKG="otelcol-contrib_0.140.0_${PLATFORM}.rpm"
-    URL="https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.140.0/\$PKG"
+    PKG="otelcol-contrib_0.151.0_${PLATFORM}.rpm"
+    URL="https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.151.0/\$PKG"
```
hugares, line 44: This change looks good, as a follow up improvement, we should check if we have a local proxy we can get this from. There were build cache features worked on recently in Konflux, I wonder if we have a cache for GH artifact

hugares (approved):
/lgtm

mshaposhnik:
> Only comment is we better squash these commits when merging so history is a bit cleaner

It is automated :)
image

meyrevived (approved):
/lgtm

PR #990, opened 2026-07-29
fix: add missing return err in checkIfIpIsLive to prevent nil pointer panic

### LINKED ISSUE #985: Bug: Missing error return in checkIfIpIsLive causes nil pointer panic on DNS resolution failure
## Bug Description

In `pkg/ibm/ibmz_helpers.go`, the function `checkIfIpIsLive` has a missing `return` statement after a failed `net.ResolveTCPAddr` call. When DNS resolution fails, the error is logged but execution continues to line 86, where `server.Network()` and `server.String()` are called on a nil `*net.TCPAddr`, causing a panic.

## Location

**File:** `pkg/ibm/ibmz_helpers.go`  
**Function:** `checkIfIpIsLive` (line 78)  
**Bug location:** lines 83-86

## Current Code

```go
server, err := net.ResolveTCPAddr("tcp", ip+":22")
if err != nil {
    log.Error(err, "failed to resolve ip address")
}
conn, err := net.DialTimeout(server.Network(), server.String(), 5*time.Second)
```

## Expected Behavior

When `net.ResolveTCPAddr` returns an error, the function should return that error immediately instead of proceeding to use the nil `server` pointer.

## Impact

Any DNS resolution failure for an IBM Z or IBM Power host IP address will crash the controller with a nil pointer dereference panic rather than returning a clean error to the caller.

## Suggested Fix

Add `return err` after the error log on line 84, matching the error-handling pattern used elsewhere in the same function (see lines 89 and 93).

123456wda:
@
## Summary

Fix a nil pointer dereference panic in `checkIfIpIsLive` when DNS resolution fails.

## Problem

In `pkg/ibm/ibmz_helpers.go`, the function `checkIfIpIsLive` has a missing `return` after a failed `net.ResolveTCPAddr` call. When DNS resolution fails, the error is logged but execution continues to call `server.Network()` and `server.String()` on the nil `*net.TCPAddr`, causing a panic.

## Fix

Add a `return err` after the `log.Error` call, matching the error-handling pattern used elsewhere in the same function (lines 89 and 93).

## Test

Added `checkIfIpIsLive` test that verifies DNS resolution failure returns an error rather than panicking. The test uses `"a b c"` (hostname with spaces) which causes `net.ResolveTCPAddr` to fail deterministically without any network call (0ms).

## Verification

```
go vet ./pkg/ibm/          # passes
go test ./pkg/ibm/ -v      # new test passes
```

Fixes #985

🤖 Generated with [Claude Code](https://claude.com/claude-code)
@

### DIFF
--- a/pkg/ibm/ibmz_helpers.go
+++ b/pkg/ibm/ibmz_helpers.go
@@ -82,6 +82,7 @@ func checkIfIpIsLive(ctx context.Context, ip string) error {
 	server, err := net.ResolveTCPAddr("tcp", ip+":22")
 	if err != nil {
 		log.Error(err, "failed to resolve ip address")
+		return err
 	}
 	conn, err := net.DialTimeout(server.Network(), server.String(), 5*time.Second)
 	if err != nil {
--- a/pkg/ibm/ibmz_helpers_test.go
+++ b/pkg/ibm/ibmz_helpers_test.go
@@ -1,6 +1,7 @@
 package ibm
 
 import (
+	"context"
 	"fmt"
 	"regexp"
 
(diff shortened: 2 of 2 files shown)


mshaposhnik (approved):

mshaposhnik:
Fixed in https://github.com/konflux-ci/multi-platform-controller/pull/986

PR #1031, opened 2026-09-03
Fix stale artifact race and rename bypass in agent-files workflows

glevi-rh:
Two security fixes for agent-files-defense workflows:

## 1. Stale Artifact Race

**Problem:** Enforce workflow doesn't verify analyzed commit matches current PR head. Force-push between detect and enforce can cause stale results to label/approve wrong commit.

**Fix:**
- Freshness check: compare `workflow_run.head_sha` vs `pr.head.sha`
- Skip stale runs: don't mutate labels or post status when mismatch
- Bind status to analyzed SHA: use `workflow_run.head_sha` not re-fetched `pr.head.sha`
- Fail-closed: treat undefined freshness as error, not stale

## 2. Workflow Rename Bypass

**Problem:** `git diff --name-only` only shows destination paths. Renaming workflow to `.disabled` bypasses detection.

**Fix:**
- Detect: `--name-status` extracts old + new paths from renames
- Enforce: check `previous_filename` in GitHub API

## Testing
Manual and automated tests passed.

Related: KFLUXINFRA-4254

### DIFF
--- a/.github/workflows/agent-files-detect.yaml
+++ b/.github/workflows/agent-files-detect.yaml
@@ -86,9 +86,22 @@ jobs:
           # $1 = diff range expression (e.g. "A..B" incremental or "A...B" net)
           check_range() {
             local range="$1"
-            local changed
-            changed=$(git diff --no-ext-diff --name-only "${range}")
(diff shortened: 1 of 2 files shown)


meyrevived (approved):
/lgtm

PR #1053, opened 2026-09-15
testing disabeled agent files workflows

glevi-rh:
## Summary

Resolves: KFLUXINFRA-XXXX

## Testing

- [ ] `make fmt` passes
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] CI checks pass (go-ci, mpc-test, test-e2e)


glevi-rh:
test succeeded, even when we disable the workflow and it's a required check, we can't merge the PR because it won't run.

